PR #12, opened 2023-09-17
✨ feat(front) PLAS-16: Added Button Component to Design System

bvelitchkine:
# Context
We're in the process of reproducing the original MVP with better frameworks. The core of the extension is a form with various inputs. These inputs must be added into the design system, obviously.

# Solution

In this PR, we designed a first **button component**.

### DIFF
--- a/libs/design-system/src/atoms/buttons/buttons.tsx
+++ b/libs/design-system/src/atoms/buttons/buttons.tsx
@@ -29,19 +29,9 @@ interface ButtonProps {
   onClick?: React.MouseEventHandler<HTMLButtonElement>;
 }
 
-const fontSizeOptions = {
-  primary: 'text-5',
-  secondary: 'text-3',
-};
-
-const borderRadiusOptions = {
-  primary: 'rounded-md',
-  secondary: 'rounded',
-};
-
-const paddingOptions = {
-  primary: 'px-4 py-2',
-  secondary: 'px-2 py-1',
+const typeSpecificClasses = {
+  primary: 'px-4 py-2 rounded-md text-5',
+  secondary: 'px-2 py-1 rounded text-3',
 };
 
 const Button = ({ children, type = 'primary', className, onClick }: ButtonProps) => {
@@ -49,9 +39,7 @@ const Button = ({ children, type = 'primary', className, onClick }: ButtonProps)
     <button
       className={cn(
         'text-white bg-ocean-blue shadow-lg shadow-grey-light hover:bg-green transition-colors duration-300 ease-in-out',
-        borderRadiusOptions[type],
-        fontSizeOptions[type],
-        paddingOptions[type],
+        typeSpecificClasses[type],
         className
       )}
       onClick={onClick}>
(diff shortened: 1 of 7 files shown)


libs/design-system/src/atoms/buttons/buttons.tsx
```diff
@@ -0,0 +1,64 @@
...
+
+const borderRadiusOptions = {
+  primary: 'rounded-md',
+  secondary: 'rounded',
+};
+
+const paddingOptions = {
+  primary: 'px-4 py-2',
+  secondary: 'px-2 py-1',
+};
```
GauvainThery: Maybe slightly overengineered especially if we end up with 3 or 4 button types.

Something like:
```
const typeSpecificClasses = {
    primary: 'px-4 py-2 rounded-md text-5',
    secondary: '...'
}


[...]

      className={cn(
        'text-white bg-ocean-blue shadow-lg shadow-grey-light hover:bg-green transition-colors duration-300 ease-in-out',
        typeSpecificClasses[type],
        className
      )}
```

could be enough. What do you think?

libs/design-system/tailwind.config.js
```diff
@@ -39,6 +40,20 @@ export default {
...
+      'ocean-blue': '#00011d',
+      cyan: '#acd8ff',
+      pink: '#fcb2d9',
+      green: '#2ee59d',
+      white: '#ffffff',
+      grey: {
+        light: '#798186',
+        DEFAULT: '#212529',
+      },
+      black: '#000000',
```
GauvainThery, line 55: So finally you went for a semantic naming rather than functional (eg. primary-bg, ...)?

libs/design-system/src/atoms/buttons/buttons.tsx
```diff
@@ -0,0 +1,64 @@
+import { ReactNode } from 'react'; // createElement,
```
GauvainThery:
```suggestion
import { ReactNode } from 'react';
```

GauvainThery (approved):
So nice to review this first PR !!! 

Only a few questions and optims but nothing blocking 👌

GauvainThery:
@Bassvelitchkine One more (important) comment!!

you have to to export `export * from './buttons';` in `src/atoms/index.ts` to be able to use you buttons 😬

libs/design-system/src/atoms/buttons/buttons.tsx
bvelitchkine: I agree, let's go

Bassvelitchkine: Yes. As I went over the code again I finally thought it made more sense like this.

libs/design-system/tailwind.config.js
bvelitchkine: Yes. As I went over the code again I finally thought it made more sense like this.

GauvainThery (commented):
Let's go 🚀

PR #14, opened 2023-09-18
✨ feat(back) PLAS-23: create linkedin profile scraper

GauvainThery:
# Context

Re-creating what is called LinkedIn parser on the original project. This is only for LinkedIn, there's a dedicated issue for SalesNav. This is only the scraper, the way it'll be called is yet to be built when integrating the form.

# Solution

There's a challenge here for maintainability as LinkedIn is likely to change the structure of the front page. I tried to query elements 1 by 1 and return an empty string if not found. I created a dedicated file for data cleaning.

Also, this could have been a class but I think we'll be more efficient developing with functions and the complexity of the project is rather low so I kept things as simple as possible.

# Priority

High

### DIFF
--- a/apps/linkedin-to-notion/src/contents/linkedin-data-cleaners.ts
+++ b/apps/linkedin-to-notion/src/contents/linkedin-data-cleaners.ts
@@ -12,7 +12,7 @@ export const splitName = (fullName: string): { firstName?: string; lastName?: st
   }
   return {
     firstName: names[0].trim(),
-    lastName: names.slice(1).join(" ").trim(),
+    lastName: names.slice(1).join(' ').trim(),
   };
 };
 
--- a/apps/linkedin-to-notion/src/contents/linkedin-data-cleaners.ts
+++ b/apps/linkedin-to-notion/src/contents/linkedin-data-cleaners.ts
@@ -1,0 +1,26 @@
+/**
+ * Basic function to split first and last name in common case ie. they are split by a space
+ * @param fullName Full name eg. 'Gauvain Thery'
+ * @returns an object with a split between first and last names
+ */
+export const splitName = (fullName: string): { firstName?: string; lastName?: string } => {
+  const names = fullName.split(' ');
+  if (names.length === 1) {
+    return {
+      firstName: names[0],
+    };
+  }
+  return {
+    firstName: names[0].trim(),
+    lastName: names[names.length - 1].trim(),
+  };
+};
+
+/**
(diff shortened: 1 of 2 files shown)


apps/linkedin-to-notion/src/contents/linkedin-data-cleaners.ts
```diff
@@ -0,0 +1,26 @@
...
+export const splitName = (fullName: string): { firstName?: string; lastName?: string } => {
+  const names = fullName.split(' ');
+  if (names.length === 1) {
+    return {
+      firstName: names[0],
+    };
+  }
+  return {
+    firstName: names[0].trim(),
+    lastName: names[names.length - 1].trim(),
```
bvelitchkine: Here, you've taken the last element of `names` as the last name, but what would happen with *Jean De La Rochebrochard* for instance? That's why I think this is better:

```suggestion
    lastName: names.slice(1).join(" ").trim(),
```

apps/linkedin-to-notion/src/contents/linkedin-profile-scraper.ts
```diff
@@ -0,0 +1,126 @@
...
+    return '';
+  }
+
+  const jobTitle = lastJob.querySelector('div.display-flex.align-items-center.mr1.t-bold > span');
+  if (!jobTitle) {
+    return '';
+  }
+
+  return cleanJobTitle(jobTitle.textContent);
+};
```
bvelitchkine: I reckon you only want to get the job title from the experience's list. I agree that it's the most reliable but you might face an issue: the elements **won't be loaded in the page** when landing on it, unless you scroll farther down.

That's why I'd add the following last resort:
```
document.querySelector("div.text-body-medium.break-words").innerText.trim();
```

bvelitchkine (commented):
Nothing significant, just two tiny suggestions ;)

apps/linkedin-to-notion/src/contents/linkedin-profile-scraper.ts
GauvainThery: Indeed! I added a fallback but I still think that this can be largely optimised. I don't think it's the intent of the project but we'll have to think about reliability of this scraper and how to test it.

bvelitchkine (commented):
All good 🚀

GauvainThery:
> All good 🚀

Then you can click approve 😅

bvelitchkine (approved):
My bad aha

PR #18, opened 2023-09-21
✨ feat(front) PLAS-34: alert ds component

bvelitchkine:
# Context

A new alert component added to the design system

### Side Note

**We're gonna have a merge conflict upon merger** due to the theme: on master, we have some "gray" whereas here we have some "grey". Not a big deal though.

### DIFF
--- a/libs/design-system/src/atoms/alerts/alerts.tsx
+++ b/libs/design-system/src/atoms/alerts/alerts.tsx
@@ -1,0 +1,37 @@
+import { cn } from '../../shared/classnames';
+
+type AlertsProps = {
+  /**
+   * The type of alert to display
+   * @default 'info'
+   * But can be one of 'success' | 'info' | 'warning' | 'error'
+   * Changing the type of alerts merely changes the color of the alert
+   */
+  type: 'success' | 'info' | 'warning' | 'error';
+
+  /**
+   * The message to display in the alert
+   */
+  message: string;
+
+  /**
+   * Some additional class names
+   */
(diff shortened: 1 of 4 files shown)


libs/design-system/src/atoms/alerts/alerts.tsx
```diff
@@ -0,0 +1,37 @@
...
+
+const colorMapping = {
+  success: 'bg-green text-grey',
+  info: 'bg-cyan text-grey',
+  warning: 'bg-pink text-grey',
+  error: 'bg-red text-white',
+};
+
+const Alert = ({ type = 'info', message, className }: AlertsProps) => {
+  return <p className={cn('antialiased p-4 rounded-md', colorMapping[type], className)}>{message}</p>;
```
GauvainThery, line 31: I would have done a div with maybe one of the text base item int but honestly not a big deal

GauvainThery (approved):
LGTM!
Just one minor comment 😊

libs/design-system/src/atoms/alerts/alerts.tsx
bvelitchkine: Yeah, I also looked at [the bootstrap doc](https://getbootstrap.com/docs/4.0/components/alerts/) (since I used bootstrap for the MVP) and they seem to use a div. But since an alert will only ever encompass text and we have something less generic than a div to put text in it, I like the idea of using a paragraph. I'll go rogue and ignore your comment for once 🙃

PR #19, opened 2023-09-21
✨ feat(front) PLAS-21: Text Area DS Component

bvelitchkine:
A *text area* form entry was added to the design system.

### DIFF
--- a/libs/design-system/src/atoms/inputs/inputs.stories.tsx
+++ b/libs/design-system/src/atoms/inputs/inputs.stories.tsx
@@ -1,6 +1,6 @@
 import type { Meta, StoryFn } from '@storybook/react';
 
-import { SelectInput, TextInput } from '.';
+import { SelectInput, TextAreaInput, TextInput } from '.';
 
 const meta = {
   title: 'Atoms/Inputs',
@@ -8,7 +8,7 @@ const meta = {
     layout: 'centered',
   },
   tags: ['autodocs'],
-} satisfies Meta<typeof TextInput | typeof SelectInput>;
+} satisfies Meta<typeof TextInput | typeof SelectInput | typeof TextAreaInput>;
 
 export default meta;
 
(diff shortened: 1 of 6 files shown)


libs/design-system/src/atoms/inputs/inputs.stories.tsx
```diff
@@ -1,14 +1,14 @@
...
+import { SelectInput, TextAreaInput, TextInput } from '.';
 
 const meta = {
   title: 'Atoms/Inputs',
   parameters: {
     layout: 'centered',
   },
   tags: ['autodocs'],
-} satisfies Meta<typeof TextInput | typeof SelectInput>;
+} satisfies Meta<typeof TextInput | typeof SelectInput | typeof TextAreaInput>;
```
GauvainThery: I think we can just go for `satisfies Meta` before it becomes a mess 😅

libs/design-system/src/atoms/inputs/textAreaInputs.tsx
```diff
@@ -0,0 +1,58 @@
...
+
+  /**
+   * Some additional class names.
+   */
+  className?: string;
+
+  /**
+   * The function that handles the change of the input.
+   */
+  handleChange: React.ChangeEventHandler<HTMLTextAreaElement>;
```
GauvainThery, line 33: Maybe calling it `onChange` is less confusing, what do you think?

GauvainThery (approved):
All good! Just one suggestion on props naming 😊

libs/design-system/src/atoms/inputs/inputs.stories.tsx
bvelitchkine: Ok, not a problem to me because, tbh, I don't really know what the implications of removing it might be yet.

libs/design-system/src/atoms/inputs/textAreaInputs.tsx
bvelitchkine: I disagree with that one because I like the fact that `onClick`/`onChange` refer to the attributes of the underlying html tag, and that `handleAnything` refers to the actual handler (a function).

Ok with you?

GauvainThery: I see the point, honestly I don't have strong position on the topic so ok for me 👌 
I'll make sure to stay consistant across the repo 😉

PR #21, opened 2023-09-24
✨ feat(front) PLAS-35: Sales Navigator Profile Scraper

bvelitchkine:
# Context

The LinkedIn profile scraper had already been coded. I used the related files as templates.

# Caveats

- I can't test the scraper as I don't have a Sales Navigator license anymore lmao
- In the final payload I return a `salesNavURL`, not a `linkedinURL`.

> Why make a difference btw SalesNav URLs and LinkedIn URLs?

**For duplicates detection**:
- Even if you have a sales navigator license, you might sometimes jump to LinkedIn. Sales Nav URLs and LinkedIn URLs are always different. So, if you have both in a same Notion database, it hints that we should be more thorough when looking for duplicates and not only compare URLs.
- Same for Twitter. If our mission is to help freelance recruiters make more cash, there's no reason to stop at LinkedIn, Sales Navigator and LinkedIn Recruiter. A sourcer goes on all sorts of platforms. And since someone can have profiles on all of these platforms, it also hints that, at some point, we'll have to revamp our deduplication process.

> We can't test it on Sales Nav, we're not gonna roll that out are we?

I agree. Let's have the file there as a reminder but let's not say publicly that we support anything else than classic LinkedIn. If it happens to work on SN, great, but that's it. **What do you think?**

### DIFF
--- a/apps/linkedin-to-notion/src/contents/salesnav-profile-scraper.ts
+++ b/apps/linkedin-to-notion/src/contents/salesnav-profile-scraper.ts
@@ -1,0 +1,117 @@
+import type { PlasmoCSConfig } from 'plasmo';
+
+import { cleanJobTitle, splitName } from './linkedin-data-cleaners';
+
+export const config: PlasmoCSConfig = {
+  matches: ['https://www.linkedin.com/sales/people/*'],
+  run_at: 'document_idle',
+};
+
+type SalesNavProfileInformation = {
+  name: {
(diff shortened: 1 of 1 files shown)


GauvainThery (approved):
LGTM! But same for me, I can't really test.

100% aligned with your comment 🙌 
Let's focus on core features for now, but let's keep this somewhere.

PR #23, opened 2023-10-03
🐛 fix(back): PLAS-046: Messaging Bugs

bvelitchkine:
Fixed a few messaging bugs due to useless code in the back + tried to limit errors in the service worker console, just a bit

### DIFF
--- a/apps/linkedin-to-notion/src/background/index.ts
+++ b/apps/linkedin-to-notion/src/background/index.ts
@@ -3,14 +3,16 @@ export {};
 const linkedInURLRegex = /linkedin.com\/in\/.+/;
 
 chrome.action.onClicked.addListener((tab) => {
-  if (tab.url.match(linkedInURLRegex)) {
+  if (tab.url?.match(linkedInURLRegex)) {
     chrome.tabs.sendMessage(tab.id, 'toggleLinkedInNotionSidePanel');
   }
 });
 
-chrome.tabs.onUpdated.addListener((tabId, _, tab) => {
-  if (tab.url.match(linkedInURLRegex)) {
-    return chrome.tabs.sendMessage(tabId, 'updateLinkedInNotionSidePanel');
+chrome.tabs.onUpdated.addListener((tabId, { status }, tab) => {
+  if (status === 'complete') {
+    if (tab.url?.match(linkedInURLRegex)) {
+      return chrome.tabs.sendMessage(tabId, 'updateLinkedInNotionSidePanel');
+    }
+    return chrome.tabs.sendMessage(tabId, 'closeLinkedInNotionSidePanel');
   }
-  return chrome.tabs.sendMessage(tabId, 'closeLinkedInNotionSidePanel');
 });
--- a/apps/linkedin-to-notion/src/background/messages/init-session.ts
+++ b/apps/linkedin-to-notion/src/background/messages/init-session.ts
@@ -1,15 +1,15 @@
-import type { PlasmoMessaging } from '@plasmohq/messaging';
+// import type { PlasmoMessaging } from '@plasmohq/messaging';
 
-import { supabase } from '~core/supabase';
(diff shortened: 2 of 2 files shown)


apps/linkedin-to-notion/src/background/index.ts
```diff
@@ -3,14 +3,16 @@ export {};
 const linkedInURLRegex = /linkedin.com\/in\/.+/;
 
 chrome.action.onClicked.addListener((tab) => {
-  if (tab.url.match(linkedInURLRegex)) {
+  if (tab.url?.match(linkedInURLRegex)) {
```
GauvainThery, line 6: Tab.url cannot be undefined so for me the `?` brings no value. But feel free to share the example of what you fixed doing so 😬

apps/linkedin-to-notion/src/background/index.ts
```diff
@@ -3,14 +3,16 @@ export {};
...
     chrome.tabs.sendMessage(tab.id, 'toggleLinkedInNotionSidePanel');
   }
 });
 
-chrome.tabs.onUpdated.addListener((tabId, _, tab) => {
-  if (tab.url.match(linkedInURLRegex)) {
-    return chrome.tabs.sendMessage(tabId, 'updateLinkedInNotionSidePanel');
+chrome.tabs.onUpdated.addListener((tabId, { status }, tab) => {
+  if (status === 'complete') {
+    if (tab.url?.match(linkedInURLRegex)) {
```
GauvainThery, line 13: Same here

apps/linkedin-to-notion/src/background/index.ts
```diff
@@ -3,14 +3,16 @@ export {};
...
+  if (tab.url?.match(linkedInURLRegex)) {
     chrome.tabs.sendMessage(tab.id, 'toggleLinkedInNotionSidePanel');
   }
 });
 
-chrome.tabs.onUpdated.addListener((tabId, _, tab) => {
-  if (tab.url.match(linkedInURLRegex)) {
-    return chrome.tabs.sendMessage(tabId, 'updateLinkedInNotionSidePanel');
+chrome.tabs.onUpdated.addListener((tabId, { status }, tab) => {
+  if (status === 'complete') {
```
GauvainThery, line 12: Good addition I think 🆗

GauvainThery (changes requested):
I don't understand the purpose of the PR 🙈 

I think we have some errors in the service worker for those reasons:
- We have files in the contents folder that don't have a `PlasmoConfig` object so the extension spreads out of LinkedIn boundaries. Not urgent IMHO ➡️ can be fixed latter
- Sometime the update of the form is not done when going on a profile for the first time (or going on a profile then to another page and coming back on a profile) but this could go in a separated issue that would also allow to use pageHistory navigation to refresh the values in the form

TL;DR
Please give me more context as I don't really see an issue and show me in what sense your fix are fixing it 😬

apps/linkedin-to-notion/src/background/index.ts
bvelitchkine: Yeah you're right. You already told me that but once again I got confused, my bad.

bvelitchkine:
~Like I said in a comment above, I got confused. Just forget this PR!~

What should we do with this one? You seem to agree that the question mark has a purpose (cf. [your comment](https://github.com/Bastien-and-Gauvain/monorepo/pull/24#discussion_r1348782392) in this PR), when I seemed to have forgotten myself aha

GauvainThery:
> ~Like I said in a comment above, I got confused. Just forget this PR!~
> 
> What should we do with this one? You seem to agree that the question mark has a purpose (cf. [your comment](https://github.com/Bastien-and-Gauvain/monorepo/pull/24#discussion_r1348782392) in this PR), when I seemed to have forgotten myself aha

Just close the PR, we'll fix that elsewhere 👌

PR #24, opened 2023-10-04
🎨 improvement(front) PLAS-047: Handle 'Go Back' in LinkedIn

bvelitchkine:
# Problem
image
Here's what used to happen:
1. We'd visit a first profile on LinkedIn
2. The data would be scraped and fill the extension fields
3. We'd visit another profile on LinkedIn
4. The data would - once again - be scraped and fill the extension fields.
5. Then we'd click on 'go back' in the browser to roll back to the first profile (cf. pink highlight on screenshot above)
6. Its data would **not be scraped**

# Solution
It's not the cleanest of solutions but it's the only one I found: a mere `setTimeout` of 2s... Anyway, that works seemlessly.

### DIFF
--- a/apps/linkedin-to-notion/src/background/index.ts
+++ b/apps/linkedin-to-notion/src/background/index.ts
@@ -9,7 +9,7 @@ chrome.action.onClicked.addListener((tab) => {
 });
 
 chrome.tabs.onUpdated.addListener((tabId, { status }, tab) => {
-  if (tab.url?.match(linkedInURLRegex)) {
+  if (tab.url && tab.url.match(linkedInURLRegex)) {
     if (status === 'complete') {
       chrome.tabs.sendMessage(tabId, 'updateLinkedInNotionSidePanel');
     }
--- a/apps/linkedin-to-notion/src/background/index.ts
+++ b/apps/linkedin-to-notion/src/background/index.ts
@@ -2,18 +2,18 @@ export {};
 
 const linkedInURLRegex = /linkedin.com\/in\/.+/;
 
-chrome.action.onClicked.addListener((tab) => {
+chrome.action.onClicked.addListener(async (tab) => {
   if (tab.url.match(linkedInURLRegex)) {
-    chrome.tabs.sendMessage(tab.id, 'toggleLinkedInNotionSidePanel');
+    return await chrome.tabs.sendMessage(tab.id, 'toggleLinkedInNotionSidePanel');
   }
 });
 
-chrome.tabs.onUpdated.addListener((tabId, { status }, tab) => {
+chrome.tabs.onUpdated.addListener(async (tabId, { status }, tab) => {
   if (tab.url && tab.url.match(linkedInURLRegex)) {
     if (status === 'complete') {
-      chrome.tabs.sendMessage(tabId, 'updateLinkedInNotionSidePanel');
+      return await chrome.tabs.sendMessage(tabId, 'updateLinkedInNotionSidePanel');
     }
   } else {
-    chrome.tabs.sendMessage(tabId, 'closeLinkedInNotionSidePanel');
+    return await chrome.tabs.sendMessage(tabId, 'closeLinkedInNotionSidePanel');
   }
 });
--- a/apps/linkedin-to-notion/src/background/index.ts
+++ b/apps/linkedin-to-notion/src/background/index.ts
@@ -8,9 +8,12 @@ chrome.action.onClicked.addListener((tab) => {
   }
 });
 
-chrome.tabs.onUpdated.addListener((tabId, _, tab) => {
-  if (tab.url.match(linkedInURLRegex)) {
-    return chrome.tabs.sendMessage(tabId, 'updateLinkedInNotionSidePanel');
+chrome.tabs.onUpdated.addListener((tabId, { status }, tab) => {
+  if (tab.url?.match(linkedInURLRegex)) {
+    if (status === 'complete') {
+      chrome.tabs.sendMessage(tabId, 'updateLinkedInNotionSidePanel');
+    }
+  } else {
+    chrome.tabs.sendMessage(tabId, 'closeLinkedInNotionSidePanel');
   }
-  return chrome.tabs.sendMessage(tabId, 'closeLinkedInNotionSidePanel');
 });
--- a/apps/linkedin-to-notion/src/components/LinkedInNotionSidePanel/LinkedInNotionSidePanelContent.tsx
+++ b/apps/linkedin-to-notion/src/components/LinkedInNotionSidePanel/LinkedInNotionSidePanelContent.tsx
@@ -39,7 +39,8 @@ export const LinkedInNotionSidePanelContent = ({
   chrome.runtime.onMessage.addListener((msg) => {
     if (msg === 'updateLinkedInNotionSidePanel') {
       setLinkedInProfileInformation(null);
-      injectLinkedInInformation();
+      // We set a timeout to wait for the page to be fully loaded before attempting to scrape the information
+      setTimeout(() => injectLinkedInInformation(), 2000);
     }
   });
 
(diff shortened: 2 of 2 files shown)


apps/linkedin-to-notion/src/background/index.ts
```diff
@@ -8,9 +8,12 @@ chrome.action.onClicked.addListener((tab) => {
...
-  if (tab.url.match(linkedInURLRegex)) {
-    return chrome.tabs.sendMessage(tabId, 'updateLinkedInNotionSidePanel');
+chrome.tabs.onUpdated.addListener((tabId, { status }, tab) => {
+  if (tab.url?.match(linkedInURLRegex)) {
+    if (status === 'complete') {
+      chrome.tabs.sendMessage(tabId, 'updateLinkedInNotionSidePanel');
+    }
+  } else {
+    chrome.tabs.sendMessage(tabId, 'closeLinkedInNotionSidePanel');
   }
```
GauvainThery: ~Still don't think `?` is needed~

~And not a big fan of the syntax. I'd rather use returns 😬~

GauvainThery: I'd do this:
```

chrome.tabs.onUpdated.addListener((tabId, { status }, tab) => {
  if (!tab.url.match(linkedInURLRegex)) {
    return chrome.tabs.sendMessage(tabId, 'closeLinkedInNotionSidePanel');
  }

  if (status === 'complete') {
    return chrome.tabs.sendMessage(tabId, 'updateLinkedInNotionSidePanel');
  }
});

```

GauvainThery: My bad ! I found why you needed ?

But go for `if (tab.url && tab.url.match(linkedInURLRegex)) {` instead

GauvainThery: Or in my example, the ? is just fine
```
chrome.tabs.onUpdated.addListener((tabId, { status }, tab) => {
  if (!tab.url?.match(linkedInURLRegex)) {
    return chrome.tabs.sendMessage(tabId, 'closeLinkedInNotionSidePanel');
  }

  if (status === 'complete') {
    return chrome.tabs.sendMessage(tabId, 'updateLinkedInNotionSidePanel');
  }
});
```

apps/linkedin-to-notion/src/components/LinkedInNotionSidePanel/LinkedInNotionSidePanelContent.tsx
```diff
@@ -39,7 +39,8 @@ export const LinkedInNotionSidePanelContent = ({
   chrome.runtime.onMessage.addListener((msg) => {
     if (msg === 'updateLinkedInNotionSidePanel') {
       setLinkedInProfileInformation(null);
-      injectLinkedInInformation();
+      // We set a timeout to wait for the page to be fully loaded before attempting to scrape the information
+      setTimeout(() => injectLinkedInInformation(), 2000);
```
GauvainThery: For me the problem is that even though we send the message onComplete state of the tab, the message is fired to soon and the scraping is done with previous data. 

I honestly don't like the idea of the timeout and if we have to go for it I'd rather go for it in this file: `apps/linkedin-to-notion/src/background/index.ts`

apps/linkedin-to-notion/src/background/index.ts
```diff
@@ -8,9 +8,12 @@ chrome.action.onClicked.addListener((tab) => {
...
-chrome.tabs.onUpdated.addListener((tabId, _, tab) => {
-  if (tab.url.match(linkedInURLRegex)) {
-    return chrome.tabs.sendMessage(tabId, 'updateLinkedInNotionSidePanel');
+chrome.tabs.onUpdated.addListener((tabId, { status }, tab) => {
+  if (tab.url?.match(linkedInURLRegex)) {
+    if (status === 'complete') {
+      chrome.tabs.sendMessage(tabId, 'updateLinkedInNotionSidePanel');
+    }
+  } else {
+    chrome.tabs.sendMessage(tabId, 'closeLinkedInNotionSidePanel');
```
GauvainThery: I think we should start by awaiting those messages, it would be cleaner
```
chrome.tabs.onUpdated.addListener(async (tabId, { status }, tab) => {
  if (tab.url?.match(linkedInURLRegex)) {
    if (status === 'complete') {
      await chrome.tabs.sendMessage(tabId, 'updateLinkedInNotionSidePanel');
    }
  } else {
    await chrome.tabs.sendMessage(tabId, 'closeLinkedInNotionSidePanel');
```

GauvainThery: And maybe a timeOut but very short one

GauvainThery (changes requested):
Sorry, my review is a bit of a mess...

My points:
- The message is sent to soon event though we're waiting for the status to be completed

I see two solutions:
1. We deal with this in the background file and add some delay - not a big fan but it works
2. We deal with this at the level of the scraper and add another scraping after a short delay. Kind of a retry that runs even in case of success to make sure the scraping is done on the most version of the page

WDYT?

apps/linkedin-to-notion/src/components/LinkedInNotionSidePanel/LinkedInNotionSidePanelContent.tsx
bvelitchkine: Yeah I really hate the idea too but I could find a better option. I've explored many (webNavigation, webRequest, window.onLoad, etc.) to no avail. I'm okay to move it in the service worker though! (and I'm more than open to a robust solution to the issue, that's for sure).

apps/linkedin-to-notion/src/background/index.ts
bvelitchkine: The problem with your suggestion is that it will try and update the side panel on every linkedin page, even if it's not a profile page, as long as the update status is "complete". Correct me if I'm wrong but the whole iframe does not unmount when it's 'closed', it only switches its css class to 'hidden'.

What about:
```suggestion
  if (tab.url && tab.url.match(linkedInURLRegex)) {
    if (status === 'complete') {
      chrome.tabs.sendMessage(tabId, 'updateLinkedInNotionSidePanel');
    }
  } else {
    chrome.tabs.sendMessage(tabId, 'closeLinkedInNotionSidePanel');
  }
```

apps/linkedin-to-notion/src/components/LinkedInNotionSidePanel/LinkedInNotionSidePanelContent.tsx
bvelitchkine: cf. comment above + adding a timeout:
```
  if (tab.url && tab.url.match(linkedInURLRegex)) {
    if (status === 'complete') {
      return setTimeout(() => chrome.tabs.sendMessage(tabId, 'updateLinkedInNotionSidePanel'), 2000);
    }
  } else {
    return chrome.tabs.sendMessage(tabId, 'closeLinkedInNotionSidePanel');
  }
```

apps/linkedin-to-notion/src/background/index.ts
bvelitchkine: cf. comment above + awaiting messages:
```
chrome.tabs.onUpdated.addListener(async (tabId, { status }, tab) => {
  if (tab.url && tab.url.match(linkedInURLRegex)) {
    if (status === 'complete') {
      return setTimeout(() => chrome.tabs.sendMessage(tabId, 'updateLinkedInNotionSidePanel'), 2000);
    }
  } else {
    return await chrome.tabs.sendMessage(tabId, 'closeLinkedInNotionSidePanel');
  }
```

bvelitchkine:
If you agree with every comment above, I'll implement the changes ;)

GauvainThery:
> If you agree with every comment above, I'll implement the changes ;)

Next implement first and ask then for a review, it's a mess to review in the comments especially as we're modifying the same lines.

Regarding what you said, I think we're just loosing some time on stuff not so important so far so let's just carry on and merge 💪

apps/linkedin-to-notion/src/background/index.ts
GauvainThery: Ok for me but add a comment with a TODO to remind us to fix this 👌

apps/linkedin-to-notion/src/background/index.ts
GauvainThery: Fair question! Go for your solution 👌

GauvainThery (approved):
Cf. my last comment 🚀

bvelitchkine:
Couldn't put the timeout in the service worker. I had to leave it in the content script. Not a big deal though, I'll merge.

PR #25, opened 2023-10-04
✨ feat(front) PLAS-018: 'Go Back To LinkedIn' CTA

bvelitchkine:
When user tries to open the extension (by clicking on its icon) on a page that's not LinkedIn, instead of the usual side panel, we display one with a call to action to open **his own profile** on LinkedIn.

### DIFF
--- a/apps/linkedin-to-notion/src/background/index.ts
+++ b/apps/linkedin-to-notion/src/background/index.ts
@@ -2,23 +2,23 @@ export {};
 
 const linkedInURLRegex = /linkedin.com\/in\/.+/;
 
-chrome.action.onClicked.addListener((tab) => {
+chrome.action.onClicked.addListener(async (tab) => {
   if (tab.url.match(linkedInURLRegex)) {
-    chrome.tabs.sendMessage(tab.id, 'toggleLinkedInNotionSidePanel');
+    return await chrome.tabs.sendMessage(tab.id, 'toggleLinkedInNotionSidePanel');
   } else {
-    chrome.tabs.sendMessage(tab.id, 'toggleGoBackToLinkedInSidePanel');
+    return await chrome.tabs.sendMessage(tab.id, 'toggleGoBackToLinkedInSidePanel');
   }
 });
 
-chrome.tabs.onUpdated.addListener((tabId, _, tab) => {
+chrome.tabs.onUpdated.addListener(async (tabId, _, tab) => {
   if (tab.url.match(linkedInURLRegex)) {
-    return chrome.tabs.sendMessage(tabId, 'updateLinkedInNotionSidePanel');
+    return await chrome.tabs.sendMessage(tabId, 'updateLinkedInNotionSidePanel');
   }
-  return chrome.tabs.sendMessage(tabId, 'closeSidePanels');
+  return await chrome.tabs.sendMessage(tabId, 'closeSidePanels');
 });
 
-chrome.runtime.onMessage.addListener((msg) => {
+chrome.runtime.onMessage.addListener(async (msg) => {
   if (msg === 'openLinkedInTab') {
-    chrome.tabs.create({ url: 'https://www.linkedin.com/in/me/' });
+    return await chrome.tabs.create({ url: 'https://www.linkedin.com/in/me/' });
   }
 });
--- a/apps/linkedin-to-notion/src/background/index.ts
+++ b/apps/linkedin-to-notion/src/background/index.ts
@@ -5,6 +5,8 @@ const linkedInURLRegex = /linkedin.com\/in\/.+/;
 chrome.action.onClicked.addListener((tab) => {
   if (tab.url.match(linkedInURLRegex)) {
     chrome.tabs.sendMessage(tab.id, 'toggleLinkedInNotionSidePanel');
+  } else {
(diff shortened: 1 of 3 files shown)


apps/linkedin-to-notion/src/background/index.ts
```diff
@@ -5,12 +5,20 @@ const linkedInURLRegex = /linkedin.com\/in\/.+/;
...
     return chrome.tabs.sendMessage(tabId, 'updateLinkedInNotionSidePanel');
   }
-  return chrome.tabs.sendMessage(tabId, 'closeLinkedInNotionSidePanel');
+  return chrome.tabs.sendMessage(tabId, 'closeSidePanels');
+});
+
+chrome.runtime.onMessage.addListener((msg) => {
+  if (msg === 'openLinkedInTab') {
+    chrome.tabs.create({ url: 'https://www.linkedin.com/in/me/' });
+  }
```
GauvainThery: Let's await the messages from now on
```
chrome.runtime.onMessage.addListener(async (msg) => {
  if (msg === 'openLinkedInTab') {
    return await chrome.tabs.create({ url: 'https://www.linkedin.com/in/me/' });
  }
```

GauvainThery (changes requested):
For me it's either:
1. We do a popup to avoid spreading the extension on all urls and injecting code on all the pages visited by the user (I'd rather do this)
2. Or, if we do a sidepanel, it's the same side panel that just return a different content

WDYT?

apps/linkedin-to-notion/src/background/index.ts
bvelitchkine: Sure but why? I think you know something about that, that I don't.

GauvainThery: To stay on the main thread. Let's say it fails, by awaiting we know it fails otherwise it's just lost somewhere in pending state.
https://medium.com/technofunnel/javascript-async-await-c83b15950a71

bvelitchkine:
Regarding the options you mentioned, I've actually explored both:
1. If by **a** popup, you mean **THE** popup (the default one), then the problem is that the popup will always open when we click on the icon, even on LinkedIn. And by doing so, it will also disable the toggling option you've implemented. That's why I ruled this one out. If you just mean **A** popup, then, what I did is just that.
2. I started like that actually. Let's assume we went that way. Then the side panel would have not only to open on LinkedIn but everywhere else on the web. So we'd have to change [the plasmo config](https://github.com/Bastien-and-Gauvain/monorepo/blob/4e8760535e0af948a5dfe215c8a3dfa84ca9cce9/apps/linkedin-to-notion/src/contents/linkedin-notion-side-panel.tsx#L8) to not only match linkedin urls. Fair enough. But then we'd also have to change [useEffect](https://github.com/Bastien-and-Gauvain/monorepo/blob/4e8760535e0af948a5dfe215c8a3dfa84ca9cce9/apps/linkedin-to-notion/src/components/LinkedInNotionSidePanel/LinkedInNotionSidePanelContent.tsx#L34-L36) right there because scraping wouldn't work on non-linkedin pages. Finally, we'd have [more complicated rendering logic in the component](https://github.com/Bastien-and-Gauvain/monorepo/blob/4e8760535e0af948a5dfe215c8a3dfa84ca9cce9/apps/linkedin-to-notion/src/components/LinkedInNotionSidePanel/LinkedInNotionSidePanelContent.tsx#L59).

*Does that make sense?*

GauvainThery:
> Regarding the options you mentioned, I've actually explored both:
> 
> 1. If by **a** popup, you mean **THE** popup (the default one), then the problem is that the popup will always open when we click on the icon, even on LinkedIn. And by doing so, it will also disable the toggling option you've implemented. That's why I ruled this one out. If you just mean **A** popup, then, what I did is just that.
> 2. I started like that actually. Let's assume we went that way. Then the side panel would have not only to open on LinkedIn but everywhere else on the web. So we'd have to change [the plasmo config](https://github.com/Bastien-and-Gauvain/monorepo/blob/4e8760535e0af948a5dfe215c8a3dfa84ca9cce9/apps/linkedin-to-notion/src/contents/linkedin-notion-side-panel.tsx#L8) to not only match linkedin urls. Fair enough. But then we'd also have to change [useEffect](https://github.com/Bastien-and-Gauvain/monorepo/blob/4e8760535e0af948a5dfe215c8a3dfa84ca9cce9/apps/linkedin-to-notion/src/components/LinkedInNotionSidePanel/LinkedInNotionSidePanelContent.tsx#L34-L36) right there because scraping wouldn't work on non-linkedin pages. Finally, we'd have [more complicated rendering logic in the component](https://github.com/Bastien-and-Gauvain/monorepo/blob/4e8760535e0af948a5dfe215c8a3dfa84ca9cce9/apps/linkedin-to-notion/src/components/LinkedInNotionSidePanel/LinkedInNotionSidePanelContent.tsx#L59).
> 
> _Does that make sense?_

Thanks for the detailed reply! I meant indeed THE popup but I didn't know we couldn't choose what to open on the onClick of the extension icon.

I'd go with the same comment as the other PR. I think I'm being too meticulous here for things that are actually well implemented and can be change easily later. So let's go for it 🚀

GauvainThery (approved):
Cf. my last comment

PR #27, opened 2023-10-08
🎨 impr(infra) PLAS-049: Better regex that only matches when supposed to

bvelitchkine:
Regex was not specific enough and [could match too many urls](https://www.notion.so/bvelitchkine/Permissive-Regex-6d375e18ccfe4bfbaed8c0669cb3dc2a)


GauvainThery (approved):
🔥 Thanks!

PR #29, opened 2023-10-08
✨ feat(front) PLAS-040: Display 'login' button if logged out

bvelitchkine:
# Context

When the user is logged out of Notion, we don't want to display a form and 'save' button that would suggest that he could save profiles in databases. **He can't** because he's logged out.

In the former extension, this is how we checked if a user was logged in:
```
const displayLoginOrDropdown = () => {
    chrome.storage.local.get("api", ({api}) => {
        // We check if we've already performed the authentication workflow
        if (api){
          // Display some stuff
        } else {
          // Display a log in button
        }
    });
};
```
That is to say that the user was logged in iff an 'api' key was found in the local chrome storage unit.

# Solution

Even though the log in feature was not implemented yet, we'll assume that here too, a logged in user has an 'api' key in his storage. This results in three additions in the `LinkedInNotionSidePanelContent` component:
- Some conditional rendering in the component ('login' button if not logged in, form otherwise)
- A first check in `useEffect` to see if the user's logged in
- A watcher to detect storage updates

### DIFF
--- a/apps/linkedin-to-notion/src/components/LinkedInNotionSidePanel/LinkedInNotionSidePanelContent.tsx
+++ b/apps/linkedin-to-notion/src/components/LinkedInNotionSidePanel/LinkedInNotionSidePanelContent.tsx
@@ -1,6 +1,6 @@
 import logo from 'data-base64:~assets/icon.png';
 import cssText from 'data-text:~style.css';
-import { Heading2, IFramedSidePanel, Spinner } from 'design-system';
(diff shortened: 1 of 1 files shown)


GauvainThery:
Love the idea but might need some changes before I review as this is strongly dependant on [my PR to sign in using Notion](https://github.com/Bastien-and-Gauvain/monorepo/pull/30) 😬

bvelitchkine:
You're right, you did everything needed in your PR aha I'll close this one!

PR #30, opened 2023-10-08
✨ feat(front,back) PLAS-42: Login and logout to Notion

GauvainThery:
# Context

This might covers a few issues from notion:
- https://www.notion.so/bvelitchkine/Create-a-new-User-table-in-Supabase-31f95d02310c46b4badadc2dea20c93d?pvs=4
- https://www.notion.so/bvelitchkine/Create-a-new-getUser-service-in-the-user-module-063ecc92e1604245b0208e0296aedeb5?pvs=4
- https://www.notion.so/bvelitchkine/Create-a-new-signUp-service-in-the-user-module-d6ee6f2361a7482baa52b8fec71fb215?pvs=4
- https://www.notion.so/bvelitchkine/Integrate-Notion-oAuth2-0-to-be-Connect-to-Notion-button-e680ab2d321646cb846e02defcfd7800?pvs=4
- https://www.notion.so/bvelitchkine/Add-a-condition-to-verify-in-storage-if-user-is-logged-in-903b4fd0dbeb436d8ace7acbc25702e0?pvs=4

In a nutshell:
- I'm using Supabase integration to initiate
- I use the storage API of Plasmo to easily store the user in chrome storage
- I display conditionaly if the user is logged in or not
- I added the possibility to logout

# Solution

https://www.loom.com/share/a9848dc894fa40939dae2f5ac1625902?sid=bb2129a4-2ae1-4ef7-a3ea-b75435d1a55b

# Priority

High


apps/linkedin-to-notion/src/contents/linkedin-notion-side-panel.tsx
```diff
@@ -18,6 +25,45 @@ export const getStyle: PlasmoGetStyle = () => {
...
+
+  const handleOAuthLogin = async (provider: Provider, scopes = 'email') => {
+    await supabase.auth.signInWithOAuth({
+      provider,
+      options: {
+        redirectTo: window.location.href,
+        scopes,
+      },
+    });
+  };
```
bvelitchkine: Not a hundred percent sure that `scopes='email'` is useful, judging by [the doc](https://supabase.com/docs/reference/javascript/auth-signinwithoauth).

apps/linkedin-to-notion/src/background/messages/users/services/user.service.ts
```diff
@@ -0,0 +1,13 @@
+import { supabase } from '~core/supabase';
+
+export default class UserService {
```
bvelitchkine: Isn't it more a "*session* service" rather than a "user service"? It's a detail that could become confusing

apps/linkedin-to-notion/src/contents/linkedin-notion-side-panel.tsx
```diff
@@ -18,6 +25,45 @@ export const getStyle: PlasmoGetStyle = () => {
...
+    }
+
+    init();
+  }, []);
+
+  const handleOAuthLogin = async (provider: Provider, scopes = 'email') => {
+    await supabase.auth.signInWithOAuth({
+      provider,
+      options: {
+        redirectTo: window.location.href,
```
bvelitchkine: Let's assume that we want to sign up on a page with the following url: "https://www.linkedin.com/in/bastien-némésis/#".

Then, the sign up fails:
image

I think it's due to the incongruence between that url (and the resulting redirect url) and those in the authentication parameters of supabase:
image

That's why I suggest the following changes to make sure there are no parasitic trailing symbols:
```suggestion
      redirectTo: window.location.href.match(/https:\/\/[a-z]{2,3}\.linkedin\.com\/in\/[^/]+\//)[0],
```

If we follow up on the example above, the regex pattern ensures that **we remove the trailing "#" from the url** for instance.

bvelitchkine (changes requested):
There's just one change that should be made to prevent minor authenticating errors. There's a detailed comment with a (tested) suggestion ;) Apart from that, it's all good 🤙

apps/linkedin-to-notion/src/background/messages/users/services/user.service.ts
GauvainThery: Yep, I agree. I started to build it very differently but know it really looks like a session service 🤔

apps/linkedin-to-notion/src/contents/linkedin-notion-side-panel.tsx
GauvainThery: It's definitely not but I'll keep the scope as we might need it to get Notion token. Could be removed later is needed.

apps/linkedin-to-notion/src/contents/linkedin-notion-side-panel.tsx
GauvainThery: Good catch!

bvelitchkine (approved):
LGTM!

bvelitchkine:
According to the talk we had earlier this day with Gauvain ("close PRs if common sense tells you it's ok"), I'll close this one myself.

PR #31, opened 2023-10-09
🐛 fix(design,infra) PLAS-051: From Interfaces to Types to Avoid Build Errors

bvelitchkine:
Removing interfaces from the design system solved this [build problem](https://www.notion.so/bvelitchkine/Types-Not-Interfaces-e5ea5bc8f3f24c118750219d3cf2b583).


GauvainThery (approved):
Good catch 👌

PR #32, opened 2023-10-09
🐛 fix(infra) PLAS-050: Better Regex Patterns 

bvelitchkine:
On some URLs, we were trying to send messages to a non existing side panel. Which [caused errors](https://www.notion.so/bvelitchkine/Trying-To-Close-Side-Panels-in-Brave-Params-527704bad5594280b76781353e09eb8d).

This PR should solve the issue!

### DIFF
--- a/apps/linkedin-to-notion/src/background/index.ts
+++ b/apps/linkedin-to-notion/src/background/index.ts
@@ -2,24 +2,29 @@ export {};
 
 // You can test the regex here: https://regex101.com/r/RJgYar/1
 // Don't forget to update the comment when you update the regex on regex101.com
-const linkedInURLRegex = /linkedin\.com\/in\/[^/]+\/$/;
+const linkedInProfileRegex = /linkedin\.com\/in\/[^/]+\/$/;
+const linkedInRegex = /linkedin\.com/g;
 
 chrome.action.onClicked.addListener(async (tab) => {
-  if (tab.url.match(linkedInURLRegex)) {
+  if (tab.url.match(linkedInProfileRegex)) {
     return await chrome.tabs.sendMessage(tab.id, 'toggleLinkedInNotionSidePanel');
   } else {
     return await chrome.tabs.sendMessage(tab.id, 'toggleGoBackToLinkedInSidePanel');
   }
 });
 
-chrome.tabs.onUpdated.addListener(async (tabId, { status }, tab) => {
-  if (tab.url && tab.url.match(linkedInURLRegex)) {
-    if (status === 'complete') {
+chrome.tabs.onUpdated.addListener(async (tabId, { status, url }, tab) => {
+  const urlToCheck = url || tab?.url;
+  if (urlToCheck) {
+    if (urlToCheck.match(linkedInProfileRegex) && status === 'complete') {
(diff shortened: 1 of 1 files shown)


apps/linkedin-to-notion/src/background/index.ts
```diff
@@ -2,24 +2,29 @@ export {};
...
+  if (urlToCheck) {
+    if (urlToCheck.match(linkedInProfileRegex) && status === 'complete') {
       return await chrome.tabs.sendMessage(tabId, 'updateLinkedInNotionSidePanel');
     }
-  } else {
-    return await chrome.tabs.sendMessage(tabId, 'closeSidePanels');
+
+    if (!urlToCheck.match(linkedInRegex)) {
+      return await chrome.tabs.sendMessage(tabId, 'closeSidePanels');
+    }
```
GauvainThery, line 25: This means it stays open even on feed or other LinkedIn pages?
Not sure about this 🤔
I'd close the panel whenever I'm not on a profile

GauvainThery: Or you meant 
```
    if (urlToCheck.match(linkedInRegex)) {
      return await chrome.tabs.sendMessage(tabId, 'closeSidePanels');
    }
```

Which would make sense to me 😅

GauvainThery (changes requested):
Not sure about the feature we want here 🤔

apps/linkedin-to-notion/src/background/index.ts
bvelitchkine: Yeah I'm sorry, my PR is not specific enough. Let me give some additional info.

# Primary problem

**We don't want to close a panel that does not exist**. It's messy and causes errors. Fair enough.

So you'd wonder why not put something like this:
```js
if (urlToCheck) {
  if(urlToCheck.match(linkedInProfileRegex)){
    // Update side panel
  } else {
    // Close side panel
  }
}
```
But that wouldn't work properly.

# Secondary problem

A LinkedIn profile page encompasses lots of subframes which means that `chrome.tabs.onUpdate` catches lots of events. Some originate from tab urls that look like this: "https://www.linkedin.com/in/bvelitchkine/" some others, from urls that look like this: "https://www.linkedin.com/mini?some-weird-parameters".

That second url would not match against `linkedInProfileRegex`. So we'd end up closing the side panel even though we're still on a LinkedIn profile page. We don't want that either. That's why I have a secondary regex (`linkedInRegex`) that does nothing of the sort if we're still on LinkedIn.

So yeah, true: it wouldn't close the extension automatically as long as we're on LinkedIn (and if it's been opened on a profile before). But I don't find that to be much of a problem (maybe I'm wrong).

# Problems with your suggestion

- It would close the "LinkedIn Side Panel" but not the "Go Back To LinkedIn Side Panel"
- Even if the function is async and `linkedInProfileRegex` is matched against the url **before** `linkedInRegex`, I don't like the fact that a profile url **would return a match in both cases** and therefore, a "close side panel" message """could""" be triggered on a LinkedIn profile page.

## Closing notes

Maybe the primary problem isn't a big deal after all and we can leave this PR pending until we find something better.

PR #33, opened 2023-10-16
✨ feat(back) PLAS-056: Store Notion API Tokens

bvelitchkine:
# Context

We used Supabase API to login through a social provider (Notion in our case, obviously). It turns out that whilst logging into Supabase through Notion, [Supabase also yields the provider (Notion) token](https://www.notion.so/bvelitchkine/Complete-Notion-OAuth-31b439e0c79a4fc2aa81578cac88721f?pvs=4#48cb0ee0d1e74d1aba0b70970ece0703) in case we need to interact with the provider's API afterwards.

**We do need to interact with Notion's API**. So we need to store that API token for later use.

# Solution

It's simple enough, thanks to Plasmo's [storage utility](https://docs.plasmo.com/framework/storage). We just store the whole payload sent by Supabase upon authentication. Notion's API token is in there somewhere, under the `provider_token` key.

### DIFF
--- a/apps/linkedin-to-notion/package.json
+++ b/apps/linkedin-to-notion/package.json
@@ -15,7 +15,7 @@
     "@plasmohq/storage": "1.7.2",
     "@supabase/supabase-js": "2.32.0",
     "design-system": "workspace:*",
-    "plasmo": "0.82.1",
+    "plasmo": "0.83.0",
     "react": "18.2.0",
     "react-dom": "18.2.0"
   },
--- a/apps/linkedin-to-notion/package.json
+++ b/apps/linkedin-to-notion/package.json
@@ -10,13 +10,14 @@
     "package": "plasmo package"
   },
   "dependencies": {
+    "@notionhq/client": "^2.2.13",
     "@plasmohq/messaging": "0.5.0",
     "@plasmohq/storage": "1.7.2",
     "@supabase/supabase-js": "2.32.0",
+    "design-system": "workspace:*",
     "plasmo": "0.82.1",
     "react": "18.2.0",
-    "react-dom": "18.2.0",
-    "design-system": "workspace:*"
+    "react-dom": "18.2.0"
   },
   "devDependencies": {
     "@ianvs/prettier-plugin-sort-imports": "4.1.0",
--- a/apps/linkedin-to-notion/src/background/index.ts
+++ b/apps/linkedin-to-notion/src/background/index.ts
@@ -30,18 +30,15 @@ chrome.runtime.onMessage.addListener(async (msg) => {
   }
 });
 
-/**
- * A function that display changes in the secure storage on the authData key
- * Should be removed, only there for testing purposes
- */
-const temp = async () => {
+// TODO Should be removed, only there for temporary testing purposes
+const displayChangesInSecureStorage = async () => {
   const storage = new SecureStorage({ area: 'local' });
   await storage.setPassword('napoleon');
   storage.watch({
-    authData: (c) => {
+    session: (c) => {
       console.log(c.newValue);
     },
   });
 };
 
-temp();
+displayChangesInSecureStorage();
--- a/apps/linkedin-to-notion/src/background/index.ts
+++ b/apps/linkedin-to-notion/src/background/index.ts
@@ -1,3 +1,5 @@
+import { SecureStorage } from '@plasmohq/storage/secure';
+
 export {};
 
 // You can test the regex here: https://regex101.com/r/RJgYar/1
@@ -28,4 +30,18 @@ chrome.runtime.onMessage.addListener(async (msg) => {
   }
 });
 
-// test('authData');
+/**
+ * A function that display changes in the secure storage on the authData key
+ * Should be removed, only there for testing purposes
+ */
+const temp = async () => {
+  const storage = new SecureStorage({ area: 'local' });
+  await storage.setPassword('napoleon');
+  storage.watch({
+    authData: (c) => {
+      console.log(c.newValue);
+    },
+  });
+};
+
+temp();
--- a/apps/linkedin-to-notion/src/background/index.ts
+++ b/apps/linkedin-to-notion/src/background/index.ts
@@ -35,8 +35,8 @@ const displayChangesInSecureStorage = async () => {
   const storage = new SecureStorage({ area: 'local' });
   await storage.setPassword('napoleon');
   storage.watch({
-    session: (c) => {
-      console.log(c.newValue);
+    notionToken: (c) => {
+      console.log('New value of notionToken: ', c.newValue);
     },
   });
 };
--- a/apps/linkedin-to-notion/src/background/index.ts
+++ b/apps/linkedin-to-notion/src/background/index.ts
@@ -2,7 +2,7 @@ export {};
 
 // You can test the regex here: https://regex101.com/r/RJgYar/1
 // Don't forget to update the comment when you update the regex on regex101.com
-export const linkedInURLRegex = /linkedin\.com\/in\/[^/]+\/$/;
+export const linkedInURLRegex = /linkedin\.com\/in\/[^/]+\/#?$/;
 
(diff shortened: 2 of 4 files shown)


apps/linkedin-to-notion/package.json
```diff
@@ -10,13 +10,14 @@
     "package": "plasmo package"
   },
   "dependencies": {
+    "@notionhq/client": "^2.2.13",
     "@plasmohq/messaging": "0.5.0",
     "@plasmohq/storage": "1.7.2",
     "@supabase/supabase-js": "2.32.0",
-    "plasmo": "0.82.1",
+    "design-system": "workspace:*",
+    "plasmo": "0.83.0",
```
GauvainThery, line 18: You're not afraid 😅

GauvainThery: (fine with updating the version here but would be better to keep this for dedicated PR next time and make sure we know the diff between both versions)

apps/linkedin-to-notion/src/background/index.ts
```diff
@@ -27,3 +29,19 @@ chrome.runtime.onMessage.addListener(async (msg) => {
...
+  const storage = new SecureStorage({ area: 'local' });
+  await storage.setPassword('napoleon');
+  storage.watch({
+    authData: (c) => {
+      console.log(c.newValue);
+    },
+  });
+};
+
+temp();
```
GauvainThery: You can use `// TODO` to highlight the todo using this extension: https://marketplace.visualstudio.com/items?itemName=Gruntfuggly.todo-tree

Even if it's temp I'd call it `displayChangesInSecureStorage` and add more context to the console.log such as `console.log("Secure storage value", c.newValue);`

apps/linkedin-to-notion/src/contents/linkedin-notion-side-panel.tsx
```diff
@@ -32,6 +33,12 @@ const LinkedinNotionSidePanel = () => {
     }),
   });
 
+  const storeEntriesSecurely = async (key: string, value: object): Promise<void> => {
+    const storage = new SecureStorage({ area: 'local' });
+    await storage.setPassword('napoleon');
```
GauvainThery, line 38: I don't understand this line 🤔

pnpm-lock.yaml
```diff
@@ -39,6 +39,9 @@ importers:
 
   apps/linkedin-to-notion:
     dependencies:
+      '@notionhq/client':
```
GauvainThery, line 42: Why removing the file?

apps/linkedin-to-notion/src/contents/linkedin-notion-side-panel.tsx
GauvainThery: This function could be on a more global util file btw

apps/linkedin-to-notion/src/contents/linkedin-notion-side-panel.tsx
```diff
@@ -42,6 +49,11 @@ const LinkedinNotionSidePanel = () => {
       }
       if (data.session) {
         setUser(data.session.user);
+        // There's one thing we're specifically interested in in the data.session object
+        // It's the provider_token (i.e notion's api access token in this case)
+        // This token is necessary to make api calls to notion elsewhere in the app
+        // There's no way to retrieve this token if we don't save it here
+        storeEntriesSecurely('authData', data.session);
```
GauvainThery: I may be missing something but this is already saved in a key called user using the hook `useStorage`. Am I missing something?

GauvainThery (changes requested):
I think we're already storing the user - maybe not in a secure way. But have a look at the hook `useStorage` we're using at login

Let me know if I'm missing something 😬

apps/linkedin-to-notion/package.json
bvelitchkine: I now realize that I might have been slightly brazen ahaha

apps/linkedin-to-notion/src/contents/linkedin-notion-side-panel.tsx
bvelitchkine: ~lmao you're absolutely right 😵 except that instead of storing `data.session.user` only, we should store `data.session` altogether (bc Notion's API token is the root). Here's the piece of code you're referring to in `linkedin-notion-side-panel.tsx`:~

```js
const LinkedinNotionSidePanel = () => {
  const [isOpen, setIsOpen] = useState(true); // set-back to false before deployment
  const [user, setUser] = useStorage<User>({
    key: 'user',
    instance: new Storage({
      area: 'local',
    }),
  });

  useEffect(() => {
    async function init() {
      const { data, error } = await supabase.auth.getSession();

      if (error) {
        console.error(error);
        return;
      }
      if (data.session) {
        setUser(data.session.user); // 👈 Let's store the whole session here, not only the user
        sendToBackground({
          name: 'sessions/resolvers/init-session',
          body: {
            refresh_token: data.session.refresh_token,
            access_token: data.session.access_token,
          },
        });
      }
    }

    init();
  }, []);
```

~Let's forget about secure storage too for now!~

---
EDIT:

Not such a bad idea after all. Here's the problem: **the provider token is available in the `data.session` object right after the oauth, but not afterwards** (as you can see by comparing screen shots below).

image
image

So, because `useEffect` calls `init` each time the component mounts, we'd be able to store the provider token right after logging in, but we'd erase it afterwards (each time the component mounts but not immediately after a login).

That being said, there's still something to change in my code to account for what I just mentioned. **We should store the session object only upon first login**.

pnpm-lock.yaml
bvelitchkine: I don't understand the question: the line you commented is green with a small '+' sign on the left indicating that I inserted (and not removed) a line. Am I missing something?

apps/linkedin-to-notion/src/contents/linkedin-notion-side-panel.tsx
bvelitchkine: cf. longer comment below: let's use the hook you'd already put in the code to store the whole session

apps/linkedin-to-notion/src/background/index.ts
bvelitchkine: Nice one, thanks! ~Useless atm (cf. longer comment below as we won't use secure storage for now) but I've installed it anyway!~

apps/linkedin-to-notion/src/contents/linkedin-notion-side-panel.tsx
bvelitchkine: For syntax matters: [read this](https://docs.plasmo.com/framework/storage#secure-storage).

For the intrinsic meaning of what I've done, here's my understanding: we're encrypting keys in the local storage, so that someone who does not have the password ('napoleon' in this case) won't be able to read what's in the storage EVER.

That being said, 'napoleon' is a hardcoded password. It's the same for everyone atm, and anyone who sneaks in our public repo would easily read the storage. I'm not sure how, but I don't see the point of a secure storage if it wasn't possible somehow.

Anyway, for the whole thing to work, I think we should have one password per user. I'm not sure what to use though. I've seen other repos using uids generated elsewhere in their app or even a password entered manually by their users. 

*Do you have any idea of what we could use as a pwd?*

---

On the other matter (making the function a util in another file): I don't want to do it now if that's ok, bc we use the function only once atm.

pnpm-lock.yaml
GauvainThery: My bad, indeed!

apps/linkedin-to-notion/src/contents/linkedin-notion-side-panel.tsx
GauvainThery: What I don't understand is that your storing securely something that is already stored elsewhere so there's no additional security. Let's talk about this IRL because I don't understand sorry 😬

GauvainThery (approved):
LGTM 🚀

PR #34, opened 2023-10-16
✨ feat(front) PLAS-055: Toggle Between LI and Notion values

bvelitchkine:
# Context

When we encounter profiles that have already been saved in the db, we need to be able to switch between values scraped on the linkedin profile and values stored in Notion.

# Solution

I created a toggle switch in the design system which controls a 'checked' state. Depending on the value of checked, we switch between LI and Notion values.

# To-do

- [x] Handle the case where `notionValues` is empty because atm, we have default values such that `notionValues` is never empty


libs/design-system/src/atoms/inputs/toggleInputs.tsx
```diff
@@ -0,0 +1,52 @@
...
...er-none after:bg-white after:shadow-[0_0px_3px_0_rgb(0_0_0_/_7%),_0_2px_2px_0_rgb(0_0_0_/_4%)] after:transition-[background-color_0.2s,transform_0.2s] after:content-[''] checked:bg-cyan checked:after:absolute checked:after:z-[2] checked:after:-mt-[3px] checked:after:ml-[1.0625rem] checked:after:h-5 checked:after:w-5 checked:after:rounded-full checked:after:border-none checked:after:bg-primary checked:after:shadow-[0_3px_1px_-2px_rgba(0,0,0,0.2),_0_2px_2px_0_rgba(0,0,0,0.14),_0_1px_5px_0_rgba(0,0,0,0.12)] checked:after:transition-[background-color_0.2s,transform_0.2s] checked:after:content-[''] hover:cursor-pointer focus:outline-none focus:ring-0 focus:before:scale-100 focus:before:opacity-[0.12] focus:before:shadow-[3px_-1px_0px_13px_rgba(0,0,0,0.6)] focus:before:transition-[box-shadow_0.2s,transform_0.2s] focus:after:absolute focus:after:z-[1] focus:after:block focus:after:h-5 focus:after:w-5 focus:after:rounded-full focus:after:content-[''] checked:focus:border-primary checked:focus:bg-primary checked:focus:before:ml-[1.0625rem] checked:focus:before:scale-100 checked:focus:before:shadow-[3px_-1px_0px_13px_#3b71ca] checked:focus:before:transition-[box-shadow_0.2s,transform_0.2s]",
```
GauvainThery: My oh my! Are you a Tailwind genius or what? 😅

apps/linkedin-to-notion/src/components/LinkedInNotionSidePanel/LinkedInNotionSidePanelContent.tsx
```diff
@@ -31,14 +32,38 @@ export const LinkedInNotionSidePanelContent = ({
...
   const [linkedInProfileInformation, setLinkedInProfileInformation] = useState<LinkedInProfileInformation | null>(null);
+  // const [notionProfileInformation, setNotionProfileInformation] = useState<NotionProfileInformation | null>(null);
 
-  const injectLinkedInInformation = async () => {
+  const setLinkedInValues = async () => {
     const scrapingResult = await getLinkedInProfileInformation();
     setLinkedInProfileInformation(scrapingResult);
   };
 
+  // const setNotionValues = async () => {
```
GauvainThery, line 47: Add a TODO comment on what todo with this and when 🙏

libs/design-system/src/atoms/inputs/selectInputs.tsx
```diff
@@ -40,7 +40,7 @@ export type SelectInputProps = {
    * The index of the initial value.
    * It's 0 by default.
    */
-  initialValue?: string;
+  selectedValue?: string;
```
GauvainThery: Why? It's the initialValue of the input before human interaction

libs/design-system/src/atoms/inputs/selectInputs.tsx
```diff
@@ -79,9 +79,9 @@ export const SelectInput = ({
       name={id}
       className={cn('antialiased border border-solid rounded-md p-2 focus:outline-none', variations[type], className)}
       onChange={handleChange}
-      defaultValue={initialValue || options[0].value}>
+      defaultValue={options[0].value}>
       {options.map(({ id, value, label }) => (
-        <option key={id} id={id} value={value}>
+        <option key={id} id={id} value={value} selected={selectedValue === value}>
```
GauvainThery: Same as above, not necessary I think

libs/design-system/src/molecules/formEntries/formEntries.stories.tsx
```diff
@@ -33,3 +35,21 @@ export const TextArea: StoryFn = () => {
...
+  const [checked, setChecked] = useState<boolean>(false);
+  const changeHandler = (event: React.ChangeEvent<HTMLInputElement>) => {
+    setChecked(event.target.checked);
+  };
+
+  return (
+    <ToggleEntry
+      inputId="toggle"
+      handleChange={changeHandler}
+      // className="w-80"
```
GauvainThery: To remove?

libs/design-system/src/atoms/inputs/toggleInputs.tsx
GauvainThery: I think you're missing `after:-ml-[0.1875rem]`  though

GauvainThery (approved):
It works and it's ok so I'll approve it as we need to deliver 🚀

I don't understand why you changed the select input though 🤔

libs/design-system/src/atoms/inputs/selectInputs.tsx
bvelitchkine: I think I need something broader than a mere initialValue because, as I switch the toggle, I also want the **status** of the profile saved in Notion to be injected in the *status entry* of the form.

statusSelectedEntryNotChanging

That being said, my solution is deprecated. According to this SO answer, [I should use `value` as well as default value](https://stackoverflow.com/a/44787318/22003549).

libs/design-system/src/atoms/inputs/selectInputs.tsx
bvelitchkine: cf. longer comment above.

libs/design-system/src/atoms/inputs/toggleInputs.tsx
bvelitchkine: Two options here: either I reveal my secret, or I keep it to myself and let you think I'm a genius 🌝

---

Yes I actually missed:
- [x] `after:-ml-[0.1875rem]`
- [x] `focus:before:shadow-[**3px**_-1px_0px_13px_rgba(0,0,0,0.6)]` ➡️  `focus:before:shadow-[**-1px**_-1px_0px_13px_rgba(0,0,0,0.6)]`

libs/design-system/src/atoms/inputs/selectInputs.tsx
GauvainThery: It's still initial according to me as you re-render the component. But I'm okay with what you did 👌

GauvainThery (approved):
LGTM 😉

PR #37, opened 2023-10-22
✨ feat(front): Prevent our style to change webpage style

### LINKED ISSUE #36: CSS is spread outside the iFrame and alters the rest of the webpage

GauvainThery:
# Context

I had some very weird changes on many pages when using the extension.

# Solution

I'm still not clear on the reason but it's related to tailwind. I applied what Plasmo does in their example, aka I used a prefix on the tailwind classes. Could be worth having something a bit more personalised. Tell me what you think! 

# Priority

High


bvelitchkine (commented):
image

Still some weird stuff going on on my end, as you can see.

However, this is all caused because of the "Go Back to LinkedIn" mumbo jumbo that requires to give the extension access to every web page. You already pointed that out awhile ago and I think you're right: we should find another. There has to be a solution!

TLDR: I'd merge this PR and open a new issue to not give the extension access to every web page.

bvelitchkine:
The TLDR of this PR aha

image

bvelitchkine (approved):
Forgot to approve

PR #39, opened 2023-10-23
🐛 fix(front): prevent the extension to spread outside LinkedIn

### LINKED ISSUE #38: Prevent the extension to spread outside LinkedIn
Currently, to display the side panel Go Back to LinkedIn, we have to make the iFrame available on all websites. This is not best practice because:
- We don't want css to spread outside the extension and by having it on all pages it increases the odds
- We don't want to add friction on other pages (loading time or anything else)

GauvainThery:
# Context

See issue.

# Solution

https://www.loom.com/share/3e06dc5a444b4f7485eb43cfa8683fa0?sid=abca9855-9dd9-46ba-a929-9cce65e74cc6
Ahah we don't see the popup in the video 😅 Pictures below ⬇️ 
image
image
image

# Priority

High


apps/linkedin-to-notion/src/popup/index.tsx
```diff
@@ -0,0 +1,43 @@
...
+      } plasmo-flex plasmo-justify-center plasmo-items-center`}>
+      {!isLinkedIn ? (
+        <GoBackToLinkedInContent />
+      ) : (
+        <BaseParagraph className="plasmo-font-bold">
+          {linkedInNotionSidePanelStatus ? 'Enjoy 🎉' : 'Goodbye 👋'}
+        </BaseParagraph>
+      )}
+    </div>
+  );
```
bvelitchkine, line 42: Here, instead of displaying the "Enjoy" and "Goodbye", I'd display **absolutely nothing**. It's uselessly distracting imo. Maybe something with a 'hidden' css class (to ensure that the popup js still executes).

bvelitchkine (approved):
Minor changes suggested, but it's not make or break, so, ok for merger!

apps/linkedin-to-notion/src/popup/index.tsx
GauvainThery: If we do so it displays an empty white square. I'll see what I can do 👌

GauvainThery: Impossible to make this square disappear and I find it very confusing...
@bvelitchkine Ok pour toi if we keep like this for now and we rediscuss the style and the wording during our styling taskforce that we'll probably do before first release?

bvelitchkine: Yeah no problem, okfm

PR #40, opened 2023-10-25
✨ feat(front, back) PLAS-053: Profile Save!

bvelitchkine:
# Solution

It's now possible to:
1. Save a LinkedIn profile in the selected Notion database
2. Open that profile directly from the extension
3. Display alerts if profile saved or error
4. And switch btw. Notion and LI values

# Testing

Don't forget to run `pnpm build` in `./libs/design-system` before running `pnpm dev` in `./apps/linkedin-to-notion`.


apps/linkedin-to-notion/src/background/messages/notion/providers/notion.provider.ts
```diff
@@ -54,4 +104,113 @@ export default class NotionProvider {
...
     return filteredEligibleDatabases;
   }
+
+  async createPageInDatabase(
+    databaseId: string,
+    linkedInProfileInformation: LinkedInProfileInformation
+  ): Promise<CreatePageResponse> {
+    const firstName = linkedInProfileInformation.name.firstName;
+    const lastName = linkedInProfileInformation.name.lastName;
+    const fullName = `${firstName} ${lastName}`;
```
GauvainThery: I'd rather create a shared.utils.ts or something like that at the base of the background folder with a function called getFullName that would take first and last names as inputs and return the full name because we'll need it for sure elsewhere

GauvainThery (approved):
Seen in sync.

PR #42, opened 2023-10-29
✨ feat (front,back) PLAS-054: Update Profile in Selected Database

bvelitchkine:
# Context

We sometimes into profiles that have already been saved in a Notion db. When that happens, we want to be warned and able to update that very profile.

# Solution

Here's what's been implemented in detail:
- [x] Check whether profile exists in pre-selected db when landing on a profile
- [x]  Check whether the profile exists in selected db as soon as selection changes
- [x]  Display an `InfoAlert` whether a profile is found in db or not
- [x]  Update the button from save to update if there are `notionValues` available (a profile was found or just saved)
- [x]  Display a  `SuccessAlert` on update


apps/linkedin-to-notion/src/background/messages/notion/providers/notion.provider.ts
```diff
@@ -3,10 +3,17 @@ import type {
...
+  UpdatePageResponse,
 } from '@notionhq/client/build/src/api-endpoints';
 
+import { getLinkedinSlug } from '~src/background/shared.utils';
+
+import {
+  databaseSearchResultsToNotionProfileInformation,
+  notionProfileInformationToNotionPage,
+  notionProfileInformationToPageProperties,
+} from '../notion.transformers';
```
GauvainThery, line 15: I'd have made one file for each but not a big deal. I don't have a strong opinion or either an argument, just a way to keep files shorter 😅

apps/linkedin-to-notion/src/background/messages/notion/providers/notion.provider.ts
```diff
@@ -88,6 +95,66 @@ export class NotionProvider {
 
     return response;
   }
+
+  async findProfileInDatabase(
```
GauvainThery, line 105:
```suggestion
  /**
   * Find a profile page in the database with a given id using its LinkedIn url
   * @param databaseId The id of the database where the page is located
   * @param linkedInUrl The LinkedIn url of the profile to find
   * @returns the response from the Notion API
   */
  async findProfileInDatabase(
```

apps/linkedin-to-notion/src/background/messages/notion/providers/notion.provider.ts
```diff
@@ -88,6 +95,66 @@ export class NotionProvider {
...
+    databaseId: string,
+    linkedInUrl: string
+  ): Promise<null | NotionProfileInformation | ErrorResponse> {
+    const slug = getLinkedinSlug(linkedInUrl);
+    if (!slug) {
+      return null;
```
GauvainThery, line 112: Maybe just a console.log saying that the slug couldn;t be extracted from the url ...

apps/linkedin-to-notion/src/background/messages/notion/providers/notion.provider.ts
```diff
@@ -88,6 +95,66 @@ export class NotionProvider {
...
+    } else {
```
GauvainThery: Ahah you and the "else" 😅
You're returning, take this out of here 🚮

apps/linkedin-to-notion/src/background/messages/notion/providers/notion.provider.ts
```diff
@@ -88,6 +95,66 @@ export class NotionProvider {
...
+
+      // The search returned a result (might be empty though)
+      if (searchResults.results.length > 0) {
+        console.log('Found smthg');
+        // We just need the first result
+        return databaseSearchResultsToNotionProfileInformation(searchResults);
+      } else {
+        console.log('Found nothing');
+        return null;
+      }
```
GauvainThery: Same as above 😬

apps/linkedin-to-notion/src/background/messages/notion/providers/notion.provider.ts
```diff
@@ -88,6 +95,66 @@ export class NotionProvider {
...
+              contains: slug,
+            },
+          },
+        });
+      } catch (error) {
+        console.error("NotionProvider, findProfileInDatabase, couldn't complete:", error);
+        return { error: "NotionProvider, findProfileInDatabase, couldn't complete:", message: JSON.stringify(error) };
+      }
+
+      // The search returned a result (might be empty though)
```
GauvainThery:
```suggestion
```

apps/linkedin-to-notion/src/background/messages/notion/providers/notion.provider.ts
```diff
@@ -88,6 +95,66 @@ export class NotionProvider {
...
+      if (searchResults.results.length > 0) {
+        console.log('Found smthg');
+        // We just need the first result
+        return databaseSearchResultsToNotionProfileInformation(searchResults);
```
GauvainThery:
```suggestion
        console.log(`A page was found in the DB ${databaseId} matching the slug ${slug}`);
        return databaseSearchResultsToNotionProfileInformation(searchResults);
```

apps/linkedin-to-notion/src/background/messages/notion/providers/notion.provider.ts
```diff
@@ -88,6 +95,66 @@ export class NotionProvider {
...
+      } else {
+        console.log('Found nothing');
+        return null;
```
GauvainThery:
```suggestion
        console.log(`No page was found in the DB ${databaseId} matching the slug ${slug}`);
        return null;
```

apps/linkedin-to-notion/src/background/messages/notion/notion.transformers.ts
```diff
@@ -0,0 +1,193 @@
...
+    lastName: properties.lastName['rich_text'][0].text.content,
+  };
+
+  const jobTitle = properties.jobTitle['rich_text'][0].text.content;
+  const company = properties.company['rich_text'][0].text.content;
+  const location = properties.location['rich_text'][0].text.content;
+  const linkedinUrl = properties.linkedinUrl['url'];
+  const status = profileStatusMap[properties.status['select'].name];
+  const gender = properties.gender['select'] ? properties.gender['select'].name : '';
+  const comment = properties.comment['rich_text'][0].text.content;
```
GauvainThery: I made a function called `getPropertyValue` on the frontend side that does this in a slightly better way, I think you can use it here

apps/linkedin-to-notion/src/components/LinkedInNotionSidePanel/Form.tsx
```diff
@@ -146,10 +180,52 @@ export const Form = ({
...
+    if ((res as unknown as ErrorResponse).error) {
+      console.log("Couldn't update the profile", res);
+      setAlertState('error');
+      return;
+    } else {
+      console.log('Profile updated in Notion', res);
+      handleNotionResponse(res);
+      setAlertState('profile-updated');
+      return;
+    }
```
GauvainThery: Else not needed

apps/linkedin-to-notion/src/components/LinkedInNotionSidePanel/NotionDatabasesSelect.tsx
```diff
@@ -7,7 +7,11 @@ import { useStorage } from '@plasmohq/storage/hook';
 
 import { getDatabaseTitle } from './utils/notionFormat.util';
 
-export const NotionDatabasesSelect = () => {
+type NotionDatabasesSelectProps = {
+  valueChangeHandler: (value: string) => void;
+};
+
+export const NotionDatabasesSelect = ({ valueChangeHandler }: NotionDatabasesSelectProps) => {
```
GauvainThery, line 14: I'd rather call it `onValueChange` but I think we alreadyhad this debate so fine for me if you want to keep it like this 😉

apps/linkedin-to-notion/src/components/LinkedInNotionSidePanel/Form.tsx
```diff
@@ -77,19 +76,53 @@ export const Form = ({
...
+        setAlertState('in-notion');
+        return;
+      }
+    } else {
+      // The search worked but no profile in selected db
+      console.log('No profile found in Notion');
+      setCurrentNotionValues(null);
+      setAlertState('new-profile');
+      return;
+    }
```
GauvainThery: Same comments on else 😬


```suggestion
    if (!res) {
        console.log('No profile found in Notion');
        setCurrentNotionValues(null);
        setAlertState('new-profile');
        return;
    }

    if ((res as unknown as ErrorResponse).error) {
        console.log("The profile query didn't work", res);
        setAlertState('error');
        setCurrentNotionValues(null);
        return;
    }

    console.log('Profile found in Notion', res);
    setCurrentNotionValues(res);
    setNotionId(res.notionId);
    setAlertState('in-notion');
    return;
```

apps/linkedin-to-notion/src/components/LinkedInNotionSidePanel/Form.tsx
```diff
@@ -77,19 +76,53 @@ export const Form = ({
       }
     };
 
-    setLinkedinUrl(window.location.href.match(/https:\/\/[a-z]{2,4}\.linkedin\.com\/in\/[^/]+\//gim)[0]);
-
     setFormValues();
   }, [displayNotionValues, linkedinValues]);
 
+  const findProfileInNotionDatabase = async (databaseId: string, linkedinUrl: string): Promise<void> => {
```
GauvainThery, line 83: I'd add an idea of Notion loading here. With a state `isNotionLoading` set to true at the beginning of the function, false at the end and render the rest of the form (underneath the DB choice) only when the loading is done, What do you think?

GauvainThery (approved):
Why not respecting the template for the name of the PR, I know it's useless but it makes some nice commit in master 😢

A few comments, suggestions and ideas for better maintainability. I approve anyway though, it works and that's what matter most for now 🚀

apps/linkedin-to-notion/src/background/messages/notion/providers/notion.provider.ts
bvelitchkine: 1. Name of the PR: my bad, I just forgot aha
2. `notion.transformers`: I like that name because I think it better conveys the idea of what goes on (i.e data transformation) which was not as clear before imo. Now that `notionProfileInformationToNotionPage` is under `notion.transformers`, we have more info than we would had we just a function and a file going by the same name.

That being said **you're the expert**, whenever something makes you feel to uncomfortable tech-wise: you should have the last say!

apps/linkedin-to-notion/src/background/messages/notion/notion.transformers.ts
bvelitchkine: Of course, stupid me! Thx

apps/linkedin-to-notion/src/components/LinkedInNotionSidePanel/Form.tsx
bvelitchkine: Agreed!

apps/linkedin-to-notion/src/background/messages/notion/providers/notion.provider.ts
GauvainThery: I'll say don't worry 😉 
No strong opinion here and I like your rationale

GauvainThery (approved):
LGTM! I merge 😉

PR #48, opened 2023-11-03
✨ feat(front,design) PLAS-063: Alerts Revamping

bvelitchkine:
# Context

New alerts in accordance with [the design system](https://www.figma.com/file/JpOzqTAvmJ1Ju6x6MZLEU8/LinkedIn-to-Notion?type=design&node-id=48-193&mode=design&t=Fsa7jQWtr01qzjsd-0).

# Solution

When put in alerts, some icons were shrunk. It's the reason why I had to add the `plasmo-shrink-0` class on all icons.

# Priority

High


libs/design-system/src/atoms/icons/svg/ArrowPath.tsx
```diff
@@ -1,11 +1,12 @@
 import { SVGIconProps } from '.';
+import { cn } from '../../..';
 
 export const ArrowPath = ({ className, size }: SVGIconProps) => (
   <svg
     xmlns="http://www.w3.org/2000/svg"
     width={size}
     height={size}
-    className={className}
+    className={cn('plasmo-shrink-0', className)}
```
GauvainThery: Not needed. I'll explain in another comment why 😉
Same for all the other ones

libs/design-system/src/molecules/alerts/alerts.tsx
```diff
@@ -0,0 +1,74 @@
...
+      className={cn(
+        'plasmo-antialiased plasmo-w-80 plasmo-h-20 plasmo-p-3 plasmo-rounded-md plasmo-flex plasmo-flex-row plasmo-items-center',
+        colorMapping[type],
+        className
+      )}>
+      <Icon type={iconsMapping[type]} className="plasmo-m-3" />
+      <p className="plasmo-m-3 plasmo-pt-[1px]">{message}</p>
+    </span>
+  );
+};
```
GauvainThery:
```suggestion
const colorMapping = {
  success: 'plasmo-bg-success plasmo-text-white-transparent90',
  info: 'plasmo-bg-info plasmo-text-white-transparent90',
  warning: 'plasmo-bg-pink plasmo-text-white-transparent90',
  error: 'plasmo-bg-error plasmo-text-white-transparent90',
};

const iconsMapping: Record<string, IconType> = {
  success: 'CheckCircle',
  info: 'InformationCircle',
  warning: 'ExclamationCircle',
  error: 'ExclamationCircle',
};

const Alert = ({ type = 'info', className, link, message }: AlertsProps) => {
  const alert = (
    <div
      className={cn(
        'plasmo-antialiased plasmo-space-x-3 plasmo-font-semibold plasmo-w-full plasmo-p-6 plasmo-leading-4 plasmo-rounded-md plasmo-flex plasmo-flex-row plasmo-items-center',
        colorMapping[type],
        className
      )}>
      <Icon type={iconsMapping[type]} />
      <p className="plasmo-flex-wrap">{message}</p>
    </div>
  );
  return link ? (
    <a href={link} target="_blank" rel="noreferrer">
      {alert}
    </a>
  ) : (
    alert
  );
};
```

big suggestion to facilitate the process but here's some explanations:
1. I don't understand why you cast IconType. Don't hesitate to use Record (ie. Record<string, IconType>)
2. Try not to not duplicate code 👉 it's easier to wrap the main element into an anchor rather than typing it twice
3. Try not to pass fixed width or height to components, we'll handle the size when using the component. And that's why your icon was in struggle. Just don't forget the wrap in this case
4. Don't forget the leading of the text 😉

GauvainThery: 5. Use space to separate elements rather than margins

GauvainThery (changes requested):
See comment 😉

libs/design-system/src/molecules/alerts/alerts.tsx
bvelitchkine: 1. Didn't know `Record`! Nice one, thank you!
2. You're right, my bad
3. Gotcha. A lot of issues could have been avoided had I used the `space` class instead of margins. Once again: my bad.
4. Duly noted.
5. Yeah of course, my bad

Thanks!

libs/design-system/src/atoms/icons/svg/ArrowPath.tsx
bvelitchkine: Tell me if I forgot something but:
1. I've committed your suggestion below (as you can see)
2. Pulled the code in my local branch
3. Fooled around with the `plasmo-shrink-0` class

Here's what happened:
image
image

Therefore (except if I'm missing smthg), we need `plasmo-shrink-0`

bvelitchkine: You know what? Let's forget about it. I suggest merging the PR as is, with my plasmo-shrinks everywhere, because, in the end, all will be replaced by the icons from `heroicons`. And as you can see, no unintended behavior with them:
image

**First one who's blocked by this pending PR can squash-merge it, as far as I'm concerned!**

GauvainThery: I removed it by putting the Icon into a div 👌 
But aligned with you!

GauvainThery (approved):
LGTM !

PR #49, opened 2023-11-03
🎨 impr(front,design) PLAS-062: New Form Entries

bvelitchkine:
# Context

The design revamping goes on. Form entries have been significantly improved. We're now able to build proper forms and complete layouts.

# Solution

Two things you should know:
1. **Some components props have changed** here and there and might've broken a view or two. I chose not to care because these layouts are about to be revamped as well.
2. I have installed a new `heroicons` package. Our icons were too difficult to resize, especially the chevron in the custom dropdown select. I believe we should replace the old SVG icons with the new ones just as well.

Therefore, don't forget to `pnpm install` the new package before running `pnpm storybook` 😉

# Priority

High


libs/design-system/src/atoms/inputs/textAreaInputs.tsx
```diff
@@ -53,7 +53,7 @@ export const TextAreaInput = ({
       id={inputId}
       name={inputId}
       className={cn(
-        'plasmo-antialiased plasmo-text-gray plasmo-border plasmo-border-solid plasmo-border-gray plasmo-rounded-md plasmo-w-full plasmo-p-2 focus:plasmo-outline-none focus:plasmo-border-cyan',
+        'plasmo-antialiased plasmo-resize-none plasmo-text-grey-light plasmo-bg-grey-transparent plasmo-border plasmo-border-solid plasmo-border-grey-transparent plasmo-rounded-md plasmo-w-full plasmo-p-2 focus:plasmo-outline-none focus:plasmo-border-main',
```
GauvainThery, line 56: Why no resize?

GauvainThery, line 56: Why no resize?

libs/design-system/src/atoms/inputs/toggleInputs.tsx
```diff
@@ -25,28 +22,42 @@ export type ToggleInputProps = {
...
-      <input
+    <div
+      className={cn(
+        className,
+        'plasmo-flex plasmo-rounded-md plasmo-flex-row plasmo-justify-between plasmo-items-center plasmo-w-full plasmo-bg-grey-transparent'
+      )}
+      id={inputId}>
+      <button
+        className={cn(
+          'plasmo-m-1 plasmo-p-2 plasmo-rounded-md plasmo-w-1/2',
```
GauvainThery: Use space in the parent component rather than margin (same below)

libs/design-system/src/atoms/inputs/selectInputs.tsx
```diff
@@ -50,46 +42,49 @@ export type SelectInputProps = {
...
   return (
-    <select
-      required={required}
-      id={id}
-      name={id}
-      className={cn(
-        'plasmo-antialiased plasmo-border plasmo-border-solid plasmo-rounded-md plasmo-p-2 focus:plasmo-outline-none',
-        variations[type],
-        className
+    <div className={cn('plasmo-relative', className)} id={id}>
```
GauvainThery, line 75: Risky to decide not to use a real inpu as if the form became more complex we won't be able to use the input API. Not a big deal if you think it's worth it to reduce complexity 👌

libs/design-system/src/atoms/inputs/toggleInputs.tsx
```diff
@@ -25,28 +22,42 @@ export type ToggleInputProps = {
...
+
+  const clickHandler = (isChecked: boolean) => {
+    setCheckedState(isChecked);
+    handleChange(isChecked);
+  };
+
   return (
-    <div className="flex space-x-4 w-36">
-      <input
+    <div
```
GauvainThery, line 39: Same here on not using native inputs - this could be a checkbox

libs/design-system/src/atoms/inputs/inputs.stories.tsx
```diff
@@ -32,54 +26,24 @@ export const SelectPrimary: StoryFn = () => {
...
-  };
+export const NewToggle: StoryFn = () => {
+  const changeHandler = (checkedStatus: boolean) => console.log(checkedStatus);
   return (
     <ToggleInput
       options={{ checked: 'Notion', unchecked: 'LinkedIn' }}
-      checked={checked}
       inputId="linkedinNotionSwitch"
       handleChange={changeHandler}
     />
```
GauvainThery:
```suggestion
    <div className="plasmo-w-60">
      <ToggleInput
        options={{ checked: 'Notion', unchecked: 'LinkedIn' }}
        inputId="linkedinNotionSwitch"
        handleChange={changeHandler}
      />
    </div>
```

Just to see that when the size is set the buttons are the right size

libs/design-system/src/atoms/inputs/selectInputs.tsx
```diff
@@ -18,24 +21,13 @@ type SelectOptionsProps = {
...
-   * It's 'primary' by default (big and white). "Secondary" is small and blue.
-   */
-  type?: 'primary' | 'secondary';
-
   /**
    * The id of the input.
    * It's also used as the name of the input.
    * It's also used as the label of the input.
    */
   id: string;
```
GauvainThery, line 30: Why removing `required`?

libs/design-system/src/atoms/inputs/toggleInputs.tsx
```diff
@@ -25,28 +22,42 @@ export type ToggleInputProps = {
...
+      className={cn(
+        className,
+        'plasmo-flex plasmo-rounded-md plasmo-flex-row plasmo-justify-between plasmo-items-center plasmo-w-full plasmo-bg-grey-transparent'
```
GauvainThery: Missing `plasmo-border plasmo-border-solid plasmo-border-grey-transparent`

GauvainThery (changes requested):
Thanks for installing the icon package, makes a lot of sense!
A few comments here and there plus make sure to fix the other PR first.

libs/design-system/src/atoms/inputs/textAreaInputs.tsx
bvelitchkine: I find them messy, especially in an already narrow side panel.

Whatever is written in this text area's supposed to end up in a text *property* field in a Notion db. It's **not a place for long text blocks**. I don't want the resizable text area to nudge users in the wrong direction.

That being said, it's not a hard opinion. My mind can be changed!

libs/design-system/src/atoms/inputs/selectInputs.tsx
bvelitchkine: I dreaded doing this but had to... If I did, it's because I found it near impossible to properly design the dropdown select.

I'd suggest moving along, and reassess if something breaks.

Okfy?

libs/design-system/src/atoms/inputs/toggleInputs.tsx
bvelitchkine: For this one, I did not even try to format the native checkbox HTML Input. I'd already lost tons of time trying to reshape the dropdown select, so, for this one, I did not bother at all.

My suggestion's the same as before: we go for it and see later if anything breaks.

What do you think?

libs/design-system/src/atoms/inputs/selectInputs.tsx
bvelitchkine: I went too fast. I know that `required` is a default props on the native HTML select. So, since I removed that select for a custom one, I thought we could get rid of the required prop.

Now that you mention it though, I see how that's a mistake because we might still want to implement form validation and this prop could come in handy.

Will be rectified.

libs/design-system/src/atoms/inputs/toggleInputs.tsx
bvelitchkine: Yep, my bad.

bvelitchkine:
Following your suggestions, I modified my code + spotted a mismanagement of 'label' vs 'values' in my dropdown select (see related commit for additional info).

As far as I'm concerned, I think we could merge this PR as is.

libs/design-system/src/atoms/inputs/selectInputs.tsx
GauvainThery: Okpm! That's why no one never re-design the dropdown 😅 And keeps it ugly but maximise compatibility. But we're developing a product to link LinkedIn to Notion I think we're safe 😉

GauvainThery (approved):
LGTM !

PR #51, opened 2023-11-04
🎨 impr(front) PLAS-059: No 'Refresh' button on Notion values

bvelitchkine:
# Context

There's a "refresh" button to re-scrape data from the LinkedIn profile page and fill the extension's fields with them. We don't want to display this button on a profile that's already in db when we're looking at Notion values.

# Solution
noRefreshOnNotionValues

# Priority

Low


GauvainThery (approved):
Nice!

PR #53, opened 2023-11-07
✨ feat(front,design) PLAS-69: All screens implemented

GauvainThery:
# Context

- Implementation of all the different screens of the side panel
- Some fixes on the select input
- More feedback on clickable alerts
- New loader

# Solution

I'm not happy with so many things ahah 😅. Mainly
- The loading system (too many loaders spread around the app)
- The way we handle the form (we should use something like Formik or a kind of form abstraction and we should have 2 form one for LinkedIn and one for Notion)
- Some codes to refac here and there
- Notion databases shouldn't be re-fetch at every page we go to

https://www.loom.com/share/de9db58c614142c08ddb8748a5f9423a?sid=0cd489d5-60ed-4458-a5fc-914f80cad81c

# Priority

High

### DIFF
--- a/apps/linkedin-to-notion/src/components/LinkedInNotionSidePanel/NotionDatabasesSelect.tsx
+++ b/apps/linkedin-to-notion/src/components/LinkedInNotionSidePanel/NotionDatabasesSelect.tsx
@@ -73,7 +73,10 @@ export const NotionDatabasesSelect = () => {
       {!selectedNotionDatabase && (
         <>
           <ButtonPrimary onClick={notionDatabasesSetter}>Refresh databases</ButtonPrimary>
-          <ErrorAlert message="No databases found. Click here to create a template." link="https://www.notion.so/7f997332291a4e80beebecbd489430b5?v=b993efeb1b064d6d8a261fd46a8fa07f&duplicate=true" />
+          <ErrorAlert
+            message="No databases found. Click here to create a template."
+            link="https://www.notion.so/7f997332291a4e80beebecbd489430b5?v=b993efeb1b064d6d8a261fd46a8fa07f&duplicate=true"
+          />
         </>
       )}
     </>
--- a/apps/linkedin-to-notion/src/components/LinkedInNotionSidePanel/NotionDatabasesSelect.tsx
+++ b/apps/linkedin-to-notion/src/components/LinkedInNotionSidePanel/NotionDatabasesSelect.tsx
@@ -1,10 +1,11 @@
 import type { DatabaseObjectResponse } from '@notionhq/client/build/src/api-endpoints';
-import { ButtonPrimary, ErrorAlert, Loader, SelectEntry } from 'design-system';
+import { ButtonPrimary, ErrorAlert, SelectEntry } from 'design-system';
 import { useEffect, useState } from 'react';
 
 import { sendToBackground } from '@plasmohq/messaging';
 import { useStorage } from '@plasmohq/storage/hook';
 
+import { FullScreenLoader } from './FullScreenLoader';
 import { getDatabaseTitle } from './utils/notionFormat.util';
 
 export const NotionDatabasesSelect = () => {
@@ -40,11 +41,7 @@ export const NotionDatabasesSelect = () => {
   }, [notionToken, selectedNotionDatabase]);
 
   if (isLoading) {
-    return (
-      <div className="plasmo-w-full plasmo-h-32 plasmo-flex plasmo-flex-col plasmo-justify-center plasmo-items-center">
-        <Loader />
-      </div>
-    );
+    return <FullScreenLoader />;
   }
 
   return (
--- a/apps/linkedin-to-notion/src/components/LinkedInNotionSidePanel/NotionDatabasesSelect.tsx
+++ b/apps/linkedin-to-notion/src/components/LinkedInNotionSidePanel/NotionDatabasesSelect.tsx
@@ -1,5 +1,5 @@
 import type { DatabaseObjectResponse } from '@notionhq/client/build/src/api-endpoints';
-import { SelectEntry, Spinner } from 'design-system';
+import { ButtonPrimary, ErrorAlert, Loader, SelectEntry } from 'design-system';
 import { useEffect, useState } from 'react';
 
 import { sendToBackground } from '@plasmohq/messaging';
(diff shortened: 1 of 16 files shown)


apps/linkedin-to-notion/src/components/LinkedInNotionSidePanel/NotionDatabasesSelect.tsx
```diff
@@ -1,63 +1,81 @@
...
+                  label: '',
+                  value: '',
+                },
+              ]
+        }
+      />
+      {!selectedNotionDatabase && (
+        <>
+          <ButtonPrimary onClick={notionDatabasesSetter}>Refresh databases</ButtonPrimary>
+          <ErrorAlert message="No databases found. Click here to create a template." link="?" />
```
bvelitchkine:
```suggestion
          <ErrorAlert message="No databases found. Click here to create a template." link="https://www.notion.so/7f997332291a4e80beebecbd489430b5?v=b993efeb1b064d6d8a261fd46a8fa07f&duplicate=true" />
```

libs/design-system/src/molecules/alerts/alerts.tsx
```diff
@@ -42,11 +35,27 @@ const iconsMapping: Record<string, IconType> = {
...
+      link && 'hover:plasmo-bg-error-medium active:plasmo-bg-error-dark'
+    }`,
+  };
+
   const alert = (
     <div
       className={cn(
         'plasmo-antialiased plasmo-space-x-3 plasmo-font-semibold plasmo-w-full plasmo-p-6 plasmo-leading-5 plasmo-rounded-md plasmo-flex plasmo-flex-row plasmo-items-center',
         colorMapping[type],
+        link && '',
```
bvelitchkine: I don't get it. Why?

libs/design-system/src/molecules/alerts/alerts.tsx
```diff
@@ -56,7 +65,7 @@ const Alert = ({ type = 'info', className, link, message }: AlertsProps) => {
     </div>
   );
   return link ? (
-    <a href={link} target="_blank" rel="noreferrer">
+    <a href={link} target="_blank" rel="noreferrer" className=" ">
```
bvelitchkine: Why this className?

bvelitchkine (changes requested):
# Blocking stuff
## What to change
Thank you for the awesome work. I think one or two things (you've already spotted) should change though:
- I made a suggestion for the link to put in the alert when a user has no eligible db yet
- I think we should better manage [Alerts](https://www.notion.so/bvelitchkine/Better-Alerts-Management-a577564e6fc04d1a9835009382d498f9?pvs=4) and [Loaders](https://www.notion.so/bvelitchkine/Better-Loaders-Management-3eccd54225fb441d93a1643556d381de?pvs=4) before merging.

As you can see, I created issues in Notion at first, but, after all, I think these changes should be committed to the current PR.

## A friendly suggestion

It's usually not the case, but this time, I would've loved atomic commits to decipher your changes - especially those related to the display of loaders and alerts.

We both agree that they should change and, I kind of spotted where the logic happened in the code, but it's in the middle of a lot of unrelated stuff.

I didn't see how I could make relevant and precise suggestions without spending an hour on your code. That's why my issues on those topics are a bit vague.

# Non-blocking

## Non blocking for this PR

I have listed the other issues (some of which you spotted yourself) that shouldn't block this PR imo but should be dealt with before release:
- [Bottom buttons alignment on Brave](https://www.notion.so/bvelitchkine/Bottom-Buttons-Alignment-on-Brave-5bf280ef723d46d0a643bb026a2f450a?pvs=4)
- [Better Notion DB Fetch Management](https://www.notion.so/bvelitchkine/Better-Notion-db-Fetch-Management-c39a92b03cfa4dddbe3ff1c9e2fa6479?pvs=4) thanks to chrome storage.

## Non blocking for the release
An issue you mentioned and should be dealt with after the release I think. That's why I put it directly in Github (since we won't use the Notion Kanban anymore): #54

libs/design-system/src/molecules/alerts/alerts.tsx
GauvainThery: Because I was tired and forgot to remove it when testing stuff 😅

libs/design-system/src/molecules/alerts/alerts.tsx
GauvainThery: Same as below, mistake my bad

GauvainThery:
> # Blocking stuff
> ## What to change
> Thank you for the awesome work. I think one or two things (you've already spotted) should change though:
> 
> * I made a suggestion for the link to put in the alert when a user has no eligible db yet
> * I think we should better manage [Alerts](https://www.notion.so/bvelitchkine/Better-Alerts-Management-a577564e6fc04d1a9835009382d498f9?pvs=4) and [Loaders](https://www.notion.so/bvelitchkine/Better-Loaders-Management-3eccd54225fb441d93a1643556d381de?pvs=4) before merging.
> 
> As you can see, I created issues in Notion at first, but, after all, I think these changes should be committed to the current PR.
> 
> ## A friendly suggestion
> It's usually not the case, but this time, I would've loved atomic commits to decipher your changes - especially those related to the display of loaders and alerts.
> 
> We both agree that they should change and, I kind of spotted where the logic happened in the code, but it's in the middle of a lot of unrelated stuff.
> 
> I didn't see how I could make relevant and precise suggestions without spending an hour on your code. That's why my issues on those topics are a bit vague.
> 
> # Non-blocking
> ## Non blocking for this PR
> I have listed the other issues (some of which you spotted yourself) that shouldn't block this PR imo but should be dealt with before release:
> 
> * [Bottom buttons alignment on Brave](https://www.notion.so/bvelitchkine/Bottom-Buttons-Alignment-on-Brave-5bf280ef723d46d0a643bb026a2f450a?pvs=4)
> * [Better Notion DB Fetch Management](https://www.notion.so/bvelitchkine/Better-Notion-db-Fetch-Management-c39a92b03cfa4dddbe3ff1c9e2fa6479?pvs=4) thanks to chrome storage.
> 
> ## Non blocking for the release
> An issue you mentioned and should be dealt with after the release I think. That's why I put it directly in Github (since we won't use the Notion Kanban anymore): #54

Noted for the commits ✅ 

I made a quick and dirty 💩 suggestion for the loader on my last commit. Let me know what you think 😉 
image

bvelitchkine (approved):

bvelitchkine:
Works like a charm, love the way you handled the loaders actually!

PS: thanks for the atomic commits: reviewing was (almost) a pleasure.

PR #55, opened 2023-11-08
🎨 impr(front,design) PLAS-076: Better Icons

bvelitchkine:
Simply replaced custom icons with heroicons wherever possible! Not much else to say.


GauvainThery (approved):
Thanks !

PR #58, opened 2023-11-10
🎨 impr(design) PLAS-085: New Extension Icon

bvelitchkine:
We had to rename the extension because of legal issues with LinkedIn. It's been temporarily named recruitivity. So we had to change the icon for something that makes sense with this new extension name.


GauvainThery (approved):
yes!

PR #60, opened 2023-11-10
🐛 fix(back): fix linkedin scraper and notion parser

GauvainThery:
# Context

LinkedIn removed a class fromt it's profile page which puts us in a very bad situation.
And we seem to have a problem for empty values in Notion.

# Solution

See code.

# Priority

Very High


bvelitchkine (approved):
Great catch too! I've tried before and after your fix. Issue is obviously solved. Ready to merge!

Issue #61, opened 2023-11-10
Architecture Reorg


bvelitchkine:
Both Gauvain and I are dissatisfied with the current project's architecture. It's become messy, both in the front and in the back.

# Background

There should be a clear separation between: services, resolvers/controllers, providers and repositories in the background. This was the [topic of a meeting](https://www.notion.so/Service-Resolver-Controller-Provider-a966aec2160d4f9aaf215b4efa7ec41b).

# Front-end
 
There's room for improvement especially in the `Form.tsx` file that has become very long and messy. Some key action points:
- [ ] Subdivide in smaller components
- [ ] Make a better use of react hooks (useContext, useFormHook, etc.)

PR #62, opened 2023-11-10
🎨 impr(infra) PLAS-074: Burnt Useless Code and Files

bvelitchkine:
# Context
Simply removed all that's useless. Doing so, I came up with this issue: #61 (to tackle post release if you agree).

# Priority

Average


GauvainThery (approved):
LGTM !

I agree that the component folder isn't ok but don't agree with your suggestion on the ticket you created, but let's move the conversation there

PR #63, opened 2023-11-11
🐛 fix(front) PLAS-087: Changed Company Selector

bvelitchkine:
# Context

The second company was selected instead of the first:
image

# Solution

A simple css selector change.

# Priority

Very High


GauvainThery (approved):
👍

PR #64, opened 2023-11-12
🐛 fix(front,back) PLAS-88: cant-re-open-extension

GauvainThery:
# Context

I fixed a couple of things here:
1. Open is now handled using storage which allows us to autonatically reopen side panel after login or when user opens a profile in a new tab
2. Form creates bugs if open when not on a profile page which led to the iframe/app to crash. It's now prevented
3. Minor fix on select (I honestly think we shouldn't have bothered with a custom picker here, sorry)
4. Notion might be slow to detect a new DB after using the template login so I added a retry on. And another retry on the find profile as well

# Priority

Very High


bvelitchkine (approved):
Works like a charm!

PR #68, opened 2023-11-26
🎨 impr(front,back) 54: form refacto

### LINKED ISSUE #54: Form Abstraction
# Context

cf. #53: we need to better handle forms.

> The way we handle the form (we should use something like Formik or a kind of form abstraction and we should have 2 form one for LinkedIn and one for Notion)

# Priority

Post-release

GauvainThery:
# Context

I decided not to do the form abstraction as expected as it would require to change our inputs and re-think too many things for no real added for now, at least.
However, I took the time to refactor the Form component and better handle or errors from the background.
No major change when using the extension but easier too maintain (far from perfect though)

# Priority

High


bvelitchkine (approved):
LGTM (once again, thanks for the atomic commits)!

PR #82, opened 2023-11-29
✨ feat(design): Added Animated Mantra to DS

### LINKED ISSUE #73: Mantras in DS
Implement this "mantra" component in the DS:

Image

The four current mantras are:
1. Recruitment is marketing, except that you don't sell a product, but the company.
2. Recruitment velocity increases success rates.
3. Personalized outreach increases response rates. Pay attention to details.
4. Constructive feedback turns unfit candidates into ambassadors.

bvelitchkine:
A set of animated mantras displayed above a loader during the onboarding.

We want to leverage labor illusion right after users authenticated so that we're able to properly fetch the database they've just duplicated in their workspace.

However, to not have them wait for a long time doing nothing, we want to show them tips on how to become better recruiters.

Look at related issue to know more.


libs/design-system/src/molecules/mantras/mantras.tsx
```diff
@@ -0,0 +1,55 @@
...
+  const [opacity, setOpacity] = useState('plasmo-opacity-100');
+
+  useEffect(() => {
+    // Choose a new mantra every `interval` seconds
+    const intervalId = setInterval(async () => {
+      setOpacity('plasmo-opacity-0');
+      // Same duration as the transition duration
+      await new Promise((resolve) => setTimeout(resolve, 500));
+      setMantra(chooseRandomMantra());
+      setOpacity('plasmo-opacity-100');
```
GauvainThery: Try not to use changing Tailwind classes as it might not work. You should use the style property.

GauvainThery (changes requested):
This is not Design System. The typo might be if it's specific, otherwise it's just a component in the app 😬

GauvainThery (approved):
Thanks !!

PR #83, opened 2023-12-03
✨ feat(feature, front): Nudge in Notion OAuth

### LINKED ISSUE #74: Nudge in Notion OAuth
During the Notion authentication, display a nudge to have users select our template:

Image

bvelitchkine:
We want to tell users which option to select during the authentication with Notion. What seems like a minor feature actually required thorough documentation exploration. Here's how I got here.

# Shortcomings of boilerplate content scripts

I wanted to use Plasmo's API to inject content in the page. [Here's the boilerplate for content injection](https://docs.plasmo.com/framework/content-scripts-ui) and here's [the example provided that's closest to our goal](https://github.com/PlasmoHQ/examples/blob/main/with-content-scripts-ui/contents/plasmo-inline.tsx).

Unfortunately, there are three problems with this simple boilerplate.

## When to inject the nudge

Notion OAuth is a single page app but the page inside which we want to inject the nudge is the second one. Which means that we can't simply wait for the URL to match the pattern:
```js
export const config: PlasmoCSConfig = {
  matches: [
    'https://www.notion.so/install-integration?response_type=code&client_id=06ba891a-ecd0-4622-8489-6dae87349cde*',
  ],
  all_frames: false,
};
```
We also have to wait until we reach step 2 of the authentication until we inject the nudge.

### Plasmo's default anchoring

By default, [`getInlineAnchor`](https://github.com/PlasmoHQ/examples/blob/94ecd6fcd400d1408cc5fbb193db3d5e96eec668/with-content-scripts-ui/contents/plasmo-inline.tsx#L7) will inject your inline content **after** the last child of the anchor (with `elm.appendChild`).

Unfortunately, that's not what we want. Without diving too deeply into details, we want our text to be in position 1, not 2.
image

And, due to the way the app was built, there's no element in the DOM that we can select such that putting the nudge as its last child would yield the desired output. So, kind of a moot point.

Yet :)

# Solutions

## Injecting the nudge at the right time

I developed a utility function called `waitForElementToExist`. Unfortunately (once again), [the canonical solution](https://bobbyhadz.com/blog/javascript-wait-for-element-to-exist) did not suffice as there's very few distinctive CSS in Notion's web app. I had to leverage the *textContent* of the node instead of its CSS and attributes.

The solution is then used inside `getInlineAnchor` which can be asynchronous, [as stated in the doc](https://docs.plasmo.com/framework/content-scripts-ui/life-cycle#inline).

## Injecting the nudge in the right place

[As stated in the doc](https://docs.plasmo.com/framework/content-scripts-ui/life-cycle#custom-dom-mounting) it's also possible to customize the DOM mounting to override the default "appendChild".

That's what I did in `mountShadowHost`.

## Styling in the page

It's not only the nudge that we want to style but also the styling of the page the nudge's injected in. Indeed, we inject a div in a grid layout that wasn't meant to receive a new element. So we had to adapt the styling of the receiving Notion app.

However, that can't be done with the kind of content scripts we just used (Content Scripts UI they're called). These were designed to make sure the styling of the extension would not impact the page it's injected in, and vice-versa.

I reached the desired effect with a [basic content script](https://docs.plasmo.com/framework/content-scripts#adding-a-single-content-script), adding a 24px row at the top of the layout receiving the nudge:
```js
  const wrapper = document.querySelector("div[style*='grid-template-rows: 1fr 1fr;']");
  const originalStyle = wrapper.getAttribute('style');
  const newStyle = originalStyle.replace('grid-template-rows: 1fr 1fr;', 'grid-template-rows: 24px 1fr 1fr;'); // 👈 here
```


apps/linkedin-to-notion/src/contents/notion-oauth-nudge-wrapper-styling.ts
```diff
@@ -0,0 +1,19 @@
+import type { PlasmoCSConfig } from 'plasmo';
+
+import { waitForElementToExist } from './utils/utils';
+
+export const config: PlasmoCSConfig = {
+  matches: [
+    'https://www.notion.so/install-integration?response_type=code&client_id=06ba891a-ecd0-4622-8489-6dae87349cde*',
```
GauvainThery: Could you make the client_id of the notion app an env variable? 😬

apps/linkedin-to-notion/src/contents/notion-oauth-nudge-wrapper-styling.ts
```diff
@@ -0,0 +1,19 @@
...
+  ],
+};
+
+(async () => {
+  await waitForElementToExist('div', 'Use a template provided by the developer');
+  // Wait 0.1s with a promise
```
GauvainThery: Why?

apps/linkedin-to-notion/src/contents/utils/utils.ts
```diff
@@ -0,0 +1,24 @@
+const customQuerySelector = (selector: string, searchedText: string): boolean => {
```
GauvainThery: Not a big fan of the `utils/utils.ts`
I'd rather go for `shared/waitForElementToExist.ts`

GauvainThery (approved):
🤯 
Not sure I get 100% of what's going on but I trust you 👍 

Be careful of not committing keys (even if they're not secrets, it's not ideal) 

Some minor non blocking comments

apps/linkedin-to-notion/src/contents/notion-oauth-nudge-wrapper-styling.ts
bvelitchkine: Good question aha! If I remember well, it's because that's about the time I needed for the page to load properly and I didn't want to code another `waitForElement` for the `<div style="... grid-template-rows: 1fr 1fr; ...">`.

bvelitchkine: Don't know if I answer your question though

bvelitchkine: cf. call: do develop a waitForElem specific to the grid.

bvelitchkine: Ok, turns out that `waitForElementToExist` and all the subsequent shenanigans weren't warranted after all

PR #85, opened 2023-12-10
🎨 impr(front): Automatically Open Extension on LinkedIn Profiles

### LINKED ISSUE #84: Open Extension Automatically
Users have to click to open the extension, which is quite annoying in their opinion. So, here's the new intended behavior:
- [x] Open the extension automatically when the user lands on a LinkedIn profile
- [x] Close the extension automatically when the user goes out of LinkedIn
- [x] If user clicks on extension outside of LinkedIn, let them click on the current CTA to go back to LI
- [x] If the user clicks on the extension icon while on LinkedIn, it just toggles the extension
- [x] If the user clicks on the extension on LinkedIn, but not on a profile, it redirects them to a profile and opens the extension

bvelitchkine:
# Context

Users complained because the extension did not open automatically anymore which is a must-have feature for recruiters.

# Solution

I changed the extension's behavior in accordance with this issue #84 

# Priority

**VERY HIGH**


GauvainThery (approved):
Not much to say 👌
Thanks for the quick fix!

Issue #67, opened 2023-11-18
Prevent Google API "MAX_WRITE_OPERATIONS_PER_MINUTE"


GauvainThery:
When listening to tab updates, we take the risk of hitting rate limitations the Google API sendMessage. We should make sure to catch the errors to avoid the app to crash and potentially run action after complete only.

PR #88, opened 2023-12-12
✨ feat(front,back) 71: implement new onboarding

### LINKED ISSUE #71: On Install Page
Create the page to display as soon as the extension is installed (cf. [developers doc](https://developer.chrome.com/docs/extensions/reference/runtime/#event-onInstalled)). It should look like this:

Image

And it should have a developer-friendly API to modify the UI depending on the step the user reached in the onboarding. There should also be a global "onboarding" state variable.

GauvainThery:
# Context

See the issue + Figma

# Solution

https://www.loom.com/share/17fb3094bfcd4c51b14c10421484539f?sid=c9d68de5-5f66-4408-86e7-bd860a13e06a

# Priority


bvelitchkine:
Thanks for the tremendous amount of work and the Loom.

# Issues

I totally understand your frustration about issues not being detailed enough and it's my fault. However, had I to do it again, I wouldn't change a thing. Let me explain my rationale.

## Rationale

### Interdependence

Some issues depend on others. And it's only after the first has been implemented that you'll make up your mind on how to implement the second. That's the first reason why I did not want to detail issues too much at first.

### Ownership

Even if I could give a clear roadmap, I wasn't sure I'd be the one tackling said issue. And I wouldn't want to impose my roadmap, especially since you're a more experience app developer. So there's a high probability that my issues will end up being a complete waste of time, for whatever reason I just couldn't anticipate due to lack of knowledge.

### Time-efficiency

I know my tendency to spend too much time on level 2 tasks (strategizing about level 1 tasks) instead of delivering. It's also a way for me to focus on tasks that actually deliver value to the users.

## Solution

That being said, **I agree 100%** that it's way more difficult to dive into an issue and figure it out as you code rather than having it broken down before writing the code. Imo, it's the one who delves into the issue who should take upon himself to breakdown the issue beforehand.

I like the idea of having high-level issues and to break them down more specifically only when the time to implement them has come. **What do you think?**

# Code challenge

Regarding your implementation of the issue, I haven't dived in your code yet, because judging by its overarching structure and the Loom, I think I can already challenge you. I could sum it up like this:

> Why not use the local storage to store the onboarding status?

The local storage remains the best way to **share context across the extension** imo. And I fear that you've framed this specific issue like you would for a typical web app, ceaselessly exchanging data with its backend. With chrome extensions (and what I've gathered so far), a lot happens in the "front-end".

To sum it up, I think it'd be easy to:
- Let the user go through the onboarding steps and update the local storage as he completes them
- Use `useStorage` in `OnboardingPage` to somehow "listen" for status updates in the local storage and update the page accordingly
- Have an event listener in the background of the app that dispatches these status updates to Supabase so we can keep track of who went how far in the onboarding. This could be done in parallel of the rest, but the update of `OnboardingPage` would not depend on what's been stored in Supabase.

**What do you think?** Am I missing something?

GauvainThery:
Let's go! Nouvelle petite démo: https://www.loom.com/share/417bdae9b6aa42d2a8f1d27b34b48b28?sid=4a8d8182-c4c0-4da9-bb14-5c7e62b517b2

'+ RLS policies activated on Supabase ✅

apps/linkedin-to-notion/example.env
```diff
@@ -1,5 +1,10 @@
+# Frontend
+PLASMO_PUBLIC_CRX_ID=
+PLASMO_PUBLIC_CRX_PREFIX=chrome-extension://
+
 PLASMO_PUBLIC_SUPABASE_URL=
 PLASMO_PUBLIC_SUPABASE_KEY=
```
bvelitchkine, line 7:
```suggestion
PLASMO_PUBLIC_NOTION_CLIENT_ID=
```

Used in the content script responsible for adding the nudge during Notion's OAuth workflow ;)

apps/linkedin-to-notion/example.env
```diff
@@ -1,5 +1,10 @@
...
 
+# Background
 CRX_KEY=
 CRX_ID=
```
bvelitchkine, line 12: Why both `PLASMO_PUBLIC_CRX_ID` and `CRX_ID`? One for dev another for prod?

apps/linkedin-to-notion/src/background/types/supabase.ts
bvelitchkine, line 1: I hope supabase helps generate these types and interfaces 😅

apps/linkedin-to-notion/src/tabs/onboarding.tsx
```diff
@@ -0,0 +1,169 @@
...
+        },
+        Tables<'users'>
+      >({
+        name: 'users/resolvers/getOrCreateUserWithAuthenticatedUserId',
+        body: { authenticatedUserId: session.user.id },
+      });
+      setUser(userData);
+    };
+
+    if (!session || session?.expires_at > Date.now()) {
```
bvelitchkine:
```suggestion
    if (!session || session?.expires_at < Date.now()) {
```
isn't it? We want to `initSession` if the expiration date is before now I think

apps/linkedin-to-notion/src/tabs/onboarding.tsx
```diff
@@ -0,0 +1,169 @@
...
+            alt="Pin Icon"
+          />
+        )}
+        <Card
+          className="plasmo-bg-background-light"
+          title={<Heading2>3 - Pin the extension</Heading2>}
+          icon={<img src={pin_icon} alt="Pin Icon" width={150} height={150} />}
+          state={state.pinExtension}
+          callToAction={
+            <ButtonPrimary state={cardStateToButtonState[state.pinExtension]} onClick={checkIsPinned}>
```
bvelitchkine: Following up on our conversion about checking whether or not the extension is pinned, I think that once the user has reached the second step of the onboarding, we could check if the extension has been pinned every second. It's not intensive operation and it would be much less confusing imo. Wdyt?

apps/linkedin-to-notion/src/contents/linkedin-notion-side-panel.tsx
```diff
@@ -24,27 +27,39 @@ export const getStyle: PlasmoGetStyle = () => {
...
+      >({
+        name: 'users/resolvers/getOrCreateUserWithAuthenticatedUserId',
+        body: { authenticatedUserId: session.user.id },
       });
+      setUser(userData);
     };
 
-    initSession();
-  }, []);
+    if (!session || session?.expires_at > Date.now()) {
```
bvelitchkine:
```suggestion
    if (!session || session?.expires_at < Date.now()) {
```
Same, I believe the session should be initialized if the expiration date is before now.

bvelitchkine (changes requested):
Some minor changes requested if I'm correct. Otherwise, all good, and, once again, congratulations aha

apps/linkedin-to-notion/example.env
GauvainThery: It'll be added in that PR then, no need to add it here?

apps/linkedin-to-notion/example.env
GauvainThery: One to expose it to the background and the other to the content. Prefixing env var with PLASMO_PUBLIC_ makes it accessible to the content. Seeing the number of variables we have I think it's ok to have it twice 👌

apps/linkedin-to-notion/src/background/types/supabase.ts
GauvainThery: Yes it does ahah
Just a command to run -> https://supabase.com/docs/guides/api/rest/generating-types

apps/linkedin-to-notion/src/tabs/onboarding.tsx
GauvainThery: Of course 🙃

apps/linkedin-to-notion/src/tabs/onboarding.tsx
GauvainThery: I don't think so, I find it a bit confusing to have stuff happening for mysterious reason. Adding the validate button makes it very straight forward

apps/linkedin-to-notion/example.env
bvelitchkine: Understood and agreed. Thanks!

apps/linkedin-to-notion/src/tabs/onboarding.tsx
bvelitchkine: I disagree, especially since the behavior you originally expected due to my bad explanation of `chrome.userSettings`. We explicitly ask the user to pin the extension. If one second after pinning it, confettis burst on screen, I think it's straight forward that it's the extension pinning that caused it. No?

apps/linkedin-to-notion/example.env
bvelitchkine: This PR has already been merged (my fault). Should I reopen it to make that change?

apps/linkedin-to-notion/src/tabs/onboarding.tsx
```diff
@@ -0,0 +1,169 @@
...
+        },
+        Tables<'users'>
+      >({
+        name: 'users/resolvers/getOrCreateUserWithAuthenticatedUserId',
+        body: { authenticatedUserId: session.user.id },
+      });
+      setUser(userData);
+    };
+
+    if (!session || session?.expires_at < Date.now()) {
```
bvelitchkine:
```suggestion
    if (!session) {
```
It's not necessary to check the expiration date because [it's only available in pro plans](https://supabase.com/docs/guides/auth/sessions#limiting-session-lifetime-and-number-of-allowed-sessions-per-user).

Furthermore, the current condition **always resolves to true** because the "expires_at" timestamp is set in such a way that it's not a UTC timestamp (take `new Date(sessions.expires_at).toISOString()` to make sure of it) and it's always inferior to Date.now().

Because of that, we trigger an endless loop and raise a MAX_WRITE_ERROR that's already been encountered in #67 .

apps/linkedin-to-notion/src/contents/linkedin-notion-side-panel.tsx
```diff
@@ -24,27 +27,39 @@ export const getStyle: PlasmoGetStyle = () => {
...
+      >({
+        name: 'users/resolvers/getOrCreateUserWithAuthenticatedUserId',
+        body: { authenticatedUserId: session.user.id },
       });
+      setUser(userData);
     };
 
-    initSession();
-  }, []);
+    if (!session || session?.expires_at < Date.now()) {
```
bvelitchkine:
```suggestion
    if (!session) {
```

see comment on `onboarding.tsx`

bvelitchkine (approved):
Resolved everything myself bc Gauvain's overwhelmed atm

PR #90, opened 2023-12-22
✨ feat(back): Save User's LinkedIn Info in Supabase

### LINKED ISSUE #75: End of OAuth
Once the Notino auth is completed:
- [ ] ~Update the onboarding status in the dedicated page~
- [x] Redirect to https://www.linkedin.com/in/me
- [x] Send the profile info to supabase
- [x] Display the new loader (cf. #72) and random mantras (cf. #73) for an arbitrary long time
- [x] When the time is spent, query eligible databases from Notion
- [x] Display the usual form

bvelitchkine:
# Context

We want to better know our users. In that respect, we want to save their LI profile info whenever they visit their own LI profile page. More details in #75 

# Solution

Everything related to labor illusion was discarded because it's become useless thanks to Gauvain's implementation of the onboarding flow.

### DIFF
--- a/apps/linkedin-to-notion/src/background/messages/users/services/user.service.ts
+++ b/apps/linkedin-to-notion/src/background/messages/users/services/user.service.ts
@@ -1,11 +1,34 @@
 import { supabase } from '~core/supabase';
 import type { Tables } from '~src/background/types/supabase';
+import type { LinkedInProfileInformation } from '~src/contents/scrapers/linkedin-profile-scraper';
 
 export enum OnboardingStatus {
   CONNECTED_TO_NOTION = 'CONNECTED_TO_NOTION',
   FIRST_PROFILE_SAVED = 'FIRST_PROFILE_SAVED',
   EXTENSION_PINNED = 'EXTENSION_PINNED',
 }
+
+export type LinkedInProfileInformationForSupabase = {
+  first_name: string;
+  last_name: string;
+  job_title: string;
+  company_name: string;
+  location: string;
+};
+
+const transformLinkedInProfileInfoForSupabase = (
+  userLinkedInProfileInfo: LinkedInProfileInformation
+): LinkedInProfileInformationForSupabase => {
+  const { name, jobTitle, company, location } = userLinkedInProfileInfo;
+
+  return {
+    first_name: name.firstName,
+    last_name: name.lastName,
+    job_title: jobTitle,
+    company_name: company,
+    location: location,
+  };
+};
 
 export class UserService {
   constructor() {}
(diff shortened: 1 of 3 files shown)


apps/linkedin-to-notion/src/background/messages/users/services/user.service.ts
```diff
@@ -1,12 +1,35 @@
...
   EXTENSION_PINNED = 'EXTENSION_PINNED',
 }
 
+export type LinkedInProfileInformationForSupabase = {
+  first_name: string;
+  last_name: string;
+  job_title: string;
+  company_name: string;
+  location: string;
+};
```
GauvainThery, line 17: Should be defined in the Supabase type file using the Supabase command

apps/linkedin-to-notion/src/components/LinkedInNotionSidePanel/LinkedInNotionSidePanelContent.tsx
```diff
@@ -30,6 +35,7 @@ export const LinkedInNotionSidePanelContent = ({
 }) => {
   const [linkedInProfileInformation, setLinkedInProfileInformation] = useState<LinkedInProfileInformation | null>(null);
   const [isLoading, setIsLoading] = useState(false);
+  const [user] = useStorage<Tables<'users'>>('user');
```
GauvainThery, line 38: So that the typing works here as well

apps/linkedin-to-notion/src/components/LinkedInNotionSidePanel/LinkedInNotionSidePanelContent.tsx
```diff
@@ -38,10 +44,26 @@ export const LinkedInNotionSidePanelContent = ({
...
+      },
+    });
+  };
+
   useEffect(() => {
     setLinkedInValues();
   }, []);
 
+  useEffect(() => {
+    if (!!user && linkedInProfileInformation?.linkedInURL.match(/linkedin\.com\/in\/me/)) {
```
GauvainThery: I think we should check if the profile was already scraped not if the user is falsy

GauvainThery:
```suggestion
    if (user?.id && linkedInProfileInformation?.linkedInURL.match(/linkedin\.com\/in\/me/)) {
```

GauvainThery (approved):
👍

apps/linkedin-to-notion/src/background/messages/users/services/user.service.ts
bvelitchkine: Will be done in a later PR (cf. our live discussion)

PR #91, opened 2023-12-23
✨ feat(front, back): First Profile Saved is Special 👀

### LINKED ISSUE #77: Create a Special First Page in Notion
When the user saves a profile for the first time, that first saved profile must have a bit of content inside:
- [x] A callout to thank the user
- [x] A new video to detail each and every feature of the extension

The page should look something like [this](https://www.notion.so/bvelitchkine/Bastien-V-litchkine-8734a5f34d3a43e58c35efb67bafd333)

### LINKED ISSUE #79: Redirection on First Profile Saved
When a first profile is saved, with the content from #77 in it, open a new tab for the user to see that very page in Notion, opened like this:

Image

bvelitchkine:
# Context

As part of the onboarding, we want users to save a first profile and look at what the results look like. This is the crux of the matter of the onboarding - the most wanted "aha moment". That why we wanted:
1. A special first profile page with content blocks inside to thank, congratulate and educate users #77 
2. A redirection so that the user can't miss the profile he just saved #79 

# Warning

Proper onboarding material still has to be shot to show:
- [ ] How to edit a profile before saving it
- [ ] How to update a profile that's already saved in the db
- [ ] How to switch between databases
- [ ] How to add/remove databases shared with the extension

Once the video(s) are shot, we'll have to upload it on Youtube and update the code accordingly.

### DIFF
--- a/apps/linkedin-to-notion/src/background/messages/notion/resolvers/createOnboardingProfileInDatabase.ts
+++ b/apps/linkedin-to-notion/src/background/messages/notion/resolvers/createOnboardingProfileInDatabase.ts
@@ -1,0 +1,101 @@
+import type {
+  AppendBlockChildrenResponse,
+  BlockObjectRequest,
+  CreatePageResponse,
+  PageObjectResponse,
+} from '@notionhq/client/build/src/api-endpoints';
+
+import type { PlasmoMessaging } from '@plasmohq/messaging';
+
+import type { ErrorResponse, NotionProfileInformation } from '../notion.type';
+import { NotionProvider } from '../providers/notion.provider';
+
+enum OnboardingContent {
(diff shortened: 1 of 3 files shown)


apps/linkedin-to-notion/src/background/messages/notion/resolvers/createOnboardingProfileInDatabase.ts
```diff
@@ -0,0 +1,100 @@
+import type {
```
GauvainThery, line 1: Not a big deal but this logic shouldn't be here, it should be in a service 😬

GauvainThery (approved):
👍

apps/linkedin-to-notion/src/background/messages/notion/resolvers/createOnboardingProfileInDatabase.ts
bvelitchkine: Will be part of a big refacto in the future. Not a priority yet (cf. our live convo)

PR #95, opened 2023-12-23
✨ feat(back): Increment Number of Profiles Saved Counter

### LINKED ISSUE #92: Profiles Saved Counter
Increment a counter in Supabase each time a user saves a profile.

- [x] Create a new column in public supabase's 'users' table
- [x] Create a new method to increment the users counter in `user.service.ts`
- [x] Create a new resolver that calls the method above in `users/resolvers`
- [x] Call that resolver each and every time a new profile's saved by a user

bvelitchkine:
# Context

We want to keep track of the number of profiles saved by a given user.

# Solution

This has been done using a PostgresSQL remote procedure call, like [this](https://supabase.com/docs/reference/javascript/rpc).

### DIFF
--- a/apps/linkedin-to-notion/src/background/types/supabase.ts
+++ b/apps/linkedin-to-notion/src/background/types/supabase.ts
@@ -9,18 +9,36 @@ export interface Database {
           created_at: string;
           id: string;
           onboarding_status: Database['public']['Enums']['onboarding_status'] | null;
+          first_name: string | null;
+          last_name: string | null;
+          job_title: string | null;
+          company_name: string | null;
+          location: string | null;
+          number_profiles_saved: number;
         };
         Insert: {
           authenticated_user?: string | null;
           created_at?: string;
           id?: string;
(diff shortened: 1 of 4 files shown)


apps/linkedin-to-notion/src/background/types/supabase.ts
```diff
@@ -9,18 +9,36 @@ export interface Database {
           created_at: string;
           id: string;
           onboarding_status: Database['public']['Enums']['onboarding_status'] | null;
+          first_name: string | null;
+          last_name: string | null;
+          job_title: string | null;
+          company_name: string | null;
+          location: string | null;
+          number_profiles_saved: number;
```
GauvainThery: To auto-generate this file: ``` npx supabase gen types typescript --project-id "deotclulvvksaqmhgjzx" --schema public > src/background/types/supabase.ts```

GauvainThery (approved):
👍

PR #98, opened 2024-01-08
✨ feat(back,front) 96: Refac how profiles are saved and save in DB Linkedin profiles (anonymously) and notion DB structure

### LINKED ISSUE #96: Create a linkedin_profiles module that contains a service and a controller called saveOneLinkedinProfile
- [x] The goal of this issue is to stop doing logic in the front end. 

- [x] Create a module called `linkedin_profiles`
- [x] This module takes a controller called `saveOneLinkedinProfile`
- [x] This resolver called a function from the service with the same name, `saveOneLinkedinProfile`
- [x] This function takes as an input what we need from the frontend:
- [x] userId
- [x] notionInformation
- [x] linkedinInformation
- [x] Add in this function the different logic that we need: save the profile in notion, save also the profile in the DB and create a new notion DB if this one is not existing in supabase, etc...

In the meantime, don't hesitate to refac the Notion module. It should follow a provider > service pattern and remove all business logic from the provider

GauvainThery:
# Context

See this [notion](https://www.notion.so/bvelitchkine/Tracking-Kick-Off-bdbcc9c5fd524bf4b063522c1c1fcdd5), as well as the related issue regarding the refactoring.

Please test thoroughly as I made a lllllooooootttttt of critical changes!

# Priority

https://www.loom.com/share/cd5e80b02cf44ec68ef6872ea7cae405?sid=4d2e4774-5984-4620-b807-9c46f9c32adf

High


apps/linkedin-to-notion/src/components/LinkedInNotionSidePanel/Form/Form.tsx
```diff
@@ -146,60 +150,27 @@ export const Form = ({
...
+    setAlertState('profile-saved');
 
     if (user.onboarding_status === OnboardingStatus.CONNECTED_TO_NOTION) {
-      const updatedUser = await sendToBackground({
-        name: 'users/resolvers/updateOnboardingStatus',
-        body: { id: user.id, onboardingStatus: OnboardingStatus.FIRST_PROFILE_SAVED },
-      });
-      setUser(updatedUser);
-      // go back to the onboarding page
+      setUser({ ...user, onboarding_status: OnboardingStatus.FIRST_PROFILE_SAVED });
```
bvelitchkine, line 173: Aren't you doing this already in `candidateProfile.service.ts` at line 46?

bvelitchkine (approved):
## Congrats

There's one slight thing to change but it's non blocking (see my comment). I love what you've done and how much closer you brought us to [this](https://www.notion.so/bvelitchkine/Service-Resolver-Controller-Provider-a966aec2160d4f9aaf215b4efa7ec41b?pvs=4).

## One remark though

I don't like the fact that the only way to track a status change is when a user updates a profile from the extension.

Cf. the "Revenue" part of AARRR: if that's the only way to track a profile's status update in a process, then we lose insights (on how to monetize the extension, on what value we bring to our users, etc).

That's why, I'd like to see what happens on the Notion side. However, that opens a slew of questions that would significantly stretch this PR (how to update supabase, what's the trigger to update supabase, what happens if a profile is deleted from Notion, etc.).

That's why I think it's ok to keep it that way for now, but not for too long. As soon as the extension can be plugged into custom databases, and we've implemented a freemium, we come back to this issue and implement 360 funnel tracking.

Wdyt?

apps/linkedin-to-notion/src/components/LinkedInNotionSidePanel/Form/Form.tsx
GauvainThery: Not really, here I'm just keeping the state aligned with the logic in DB. It's not ideal but it was just easier to implement.

Issue #101, opened 2024-02-06
Disconnection & Automatic Toggle Bugs


bvelitchkine:
![image](https://github.com/Bastien-and-Gauvain/monorepo/assets/45038783/55acb4a7-ce18-4483-a027-2550c9c44f21)

bvelitchkine:
What causes disconnections is the (yet) unexplained disappearance of the `provider_token` key from the `data.session` object [returned by supabase on oauth](https://github.com/Bastien-and-Gauvain/monorepo/blob/d75ad0885e48104731c76abafec4f6e15dfc5f3b/apps/linkedin-to-notion/src/contents/linkedin-notion-side-panel.tsx#L38-L39) because of this check:

```ts
isLoggedIn={!!session?.provider_token}
```

Here's the payload in the local storage of the extension when it asks for a reconnection:
```json
{
    "access_token": "eyJhbGciOiJIUzI1NiIsImtpZ*****",
    "token_type": "bearer",
    "expires_in": 3600,
    "expires_at": 1707192305,
    "refresh_token": "JfL1*****",
    "user": {...}
}
```

Here's the one in the storage, right after the authentication workflow:
```json
{
    "provider_token": "secret_***",
    "access_token": "eyJhbG***",
    "expires_in": 3600,
    "expires_at": 1707195497,
    "refresh_token": "y1n1Qzw***",
    "token_type": "bearer",
    "user": {...}
}
```

To explain:
1. How does it disappear from the local storage?
2. Why is it happening now and not before? What has changed? (I looked again at the last PR in search of an answer, couldn't see any).

GauvainThery:
Having a look at what we do with sessions and login. IMHO ➡️ absolutely non-sense. We'll have to take the time to fix it. I'm still trying to understand what's best to do.

bvelitchkine:
The log outs were supposedly fixed in #103 but it turns out that they're not. Users keep getting logged out from the extension every hour.

PR #102, opened 2024-02-07
🐛 fix(front): toggle extension using react state only

bvelitchkine:
# Context

#101 : a big problem.

# Solution

Never use messaging or storage unless absolutely necessary. I think this should be our development razor in the future. That's what I did here with the toggling of the side panel. [More explantions here](https://www.loom.com/share/903ab1c01d70415aae1f274006e93152?sid=6431a1d4-3e42-4f2a-b218-eb75848e9592)

# Priority

Very High


GauvainThery:
Before having a look at the code, I disagree with the behaviour. Now it always open a /me page?
I think we should clarify what we expect from clicking the icon. I find the behaviour here super confusing.

libs/design-system/src/molecules/openExtensionBubble/openExtensionBubble.tsx
```diff
@@ -0,0 +1,13 @@
+import { ImageLogo } from '../..';
+
+export type OpenExtensionBubbleProps = {
+  onClick?: () => void;
+};
+
+export const OpenExtensionBubble = ({ onClick }: OpenExtensionBubbleProps) => {
```
GauvainThery: Rather call it ExtensionBubble as Open is the behaviour and it can differ based on what is passed in the onClick

libs/design-system/src/molecules/IFramedSidePanel/IFramedSidePanel.tsx
```diff
@@ -117,6 +124,14 @@ export const IFramedSidePanel = ({
     </>
   );
 
+  const closeContent = <OpenExtensionBubble onClick={onOpenCallback} />;
```
GauvainThery: I don't understand when this is displayed. I tried and couldn't figure it out 😬

apps/linkedin-to-notion/src/popup/index.tsx
```diff
@@ -1,44 +1,19 @@
...
 export default function Popup() {
-  const [isLinkedInProfile, setIsLinkedInProfile] = useState<boolean>(false);
-  const [linkedInNotionSidePanelStatus, setLinkedInNotionSidePanelStatus] = useState<boolean>(false);
-
-  chrome.runtime.onMessage.addListener((msg) => {
-    if (msg.name === 'linkedInNotionSidePanelStatus') {
-      setLinkedInNotionSidePanelStatus(msg.body.isOpen);
+  (async () => {
+    try {
+      await chrome.tabs.create({ active: true, url: 'https://www.linkedin.com/in/me/' });
```
GauvainThery: No need to have a popup if it only redirects to a page 😬
It could be a solution but we have to think about it first

apps/linkedin-to-notion/src/contents/linkedin-notion-side-panel.tsx
```diff
@@ -13,7 +13,7 @@ import { LinkedInNotionSidePanelContent } from '../components/LinkedInNotionSide
 import { handleOAuthLogin } from './handleOAuthLogin';
 
 export const config: PlasmoCSConfig = {
-  matches: ['https://www.linkedin.com/in/*'],
+  matches: ['https://www.linkedin.com/*'],
```
GauvainThery: So we want to have the panel always open onLinkedIn? 
I think it could be smart for different reasons (side panel closed when navigating, etc.)

GauvainThery:
Just added a few comments!
Let's see in sync tomorrow what's best to do 👌

bvelitchkine:
Call feedback:
- Extension icon opens user's linkedin profile only if not on LinkedIn
- Side panel displays "recruitivity only works on linkedin profiles - yours for instance" if not on a linkedin profile page

bvelitchkine:
@GauvainThery for some reason, can't see your comments and past review, apart from the comments above. So the only thing I did is implement what we talked about live, but maybe I missed a few things you'd have liked me to change 🤷‍♂️

libs/design-system/src/molecules/IFramedSidePanel/IFramedSidePanel.tsx
```diff
@@ -117,6 +124,14 @@ export const IFramedSidePanel = ({
...
+
+  const leftPosition = isOpen
+    ? Math.min(Math.max(10, sidePanelLeftPosition), window.innerWidth - 356)
+    : window.innerWidth - 116;
+
+  const topPosition = isOpen ? null : { top: '100px' };
```
GauvainThery: This doesn't work. If you resize the panel, the icon disappears. Actually, the problem is similar for the iFramed. You should add a listener on resize, something like:

```
    const resizeHandler = () => {
      setSidePanelLeftPosition((prev) => Math.min(prev, window.innerWidth - 356));
    };
```

To make sure it stays in the window. Careful, my example here works for downsizing but not for upsizing.

Adn TBH, I think the logic shouldn't live here. This component ends displaying something that it's not.

apps/linkedin-to-notion/src/popup/index.tsx
GauvainThery: @bvelitchkine, again the popup is useless and very confusing. IMHO, I download the extension and don't know anything about it, I'm not sure I'm able to open it 😬
I think the popup should at least indicatew how to open the sidepanel or open it, nothing less and maybe nothing more.

GauvainThery (changes requested):
Better for sure but still not okay for me. See my comments. Also, not a big fan that everytime I come back onLinkedin I have a very intrusive panel opening that I have to click on to close it. This would make me uninstall the extension...

Why do we display a sidepanel on a page that is not a profile?

Why not just having the popup explicitly say that this extension works on profiles only, and a checkbox "Open automatically whenever I open a profile"

libs/design-system/src/molecules/IFramedSidePanel/IFramedSidePanel.tsx
bvelitchkine: @GauvainThery cf. L148. Does that make sense?

libs/design-system/src/molecules/openExtensionBubble/openExtensionBubble.tsx
bvelitchkine: I actually went for `LogoBubble` which is even better I think. Thanks!

bvelitchkine:
> Better for sure but still not okay for me. See my comments. Also, not a big fan that everytime I come back onLinkedin I have a very intrusive panel opening that I have to click on to close it. This would make me uninstall the extension...
> 
> Why do we display a sidepanel on a page that is not a profile?
> 
> Why not just having the popup explicitly say that this extension works on profiles only, and a checkbox "Open automatically whenever I open a profile"

- You're right: let's open automatically on LinkedIn profiles only and close automatically only when the user leaves LinkedIn (like before). I just thought that having one less disjunction would help but you're right, from a UX perspective, it's bad.
- I agree with your comment on the popup. Let's have it display "recruitivity only works on LinkedIn profiles..." + a "...yours for instance ✨" CTA when not on a LinkedIn profile, and do nothing when we're already on a LinkedIn profile.

On that last point:
- No, I don't think the extension click should open the extension anymore. That involves messaging with content scripts and it's exactly why we had bugs. I have a strong opinion on this.
- I don't think we should have that "open automatically" checkbox either (it involves storage and extra unsolicited features users never requested).

Tell me if you agree (at least for this PR)!

libs/design-system/src/molecules/IFramedSidePanel/IFramedSidePanel.tsx
bvelitchkine: Implemented a solution to handle window resize events. However, the logic is still in the `IFramedSidePanel` component so far.

apps/linkedin-to-notion/src/contents/linkedin-notion-side-panel.tsx
bvelitchkine: No, you're right. I rolled back to the previous implementation where the content script loads on LinkedIn profiles and only closes when we leave LinkedIn.

apps/linkedin-to-notion/src/popup/index.tsx
bvelitchkine: Extension click:
- On a LinkedIn profile -> "Enjoy"
- Not on a LinkedIn profile -> "Extension only works on LinkedIn profiles... yours for instance" + CTA

bvelitchkine:
- Cursor pointer on the bubble
- Gif to explain how to open the extension instead of just "enjoy"

GauvainThery (approved):
Ok for me 👍

PR #103, opened 2024-02-11
🎨 impr(back,front): Implement better session handling, fix TS and better handle panel opening

GauvainThery:
# Context

Cf. our recent conversation.
Honestly, focus on trying and making everything works. There's still a loooot of things to improve in code. But at least, the session handler should be much more bullet proof like so.

# Priority

Very High


apps/linkedin-to-notion/core/supabase.ts
```diff
@@ -1,16 +1,88 @@
...
-  auth: {
-    storage,
-    autoRefreshToken: true,
-    persistSession: true,
-    detectSessionInUrl: true,
-  },
-});
+export const supabase = createClient(
+  process.env.PLASMO_PUBLIC_SUPABASE_URL || '',
+  process.env.PLASMO_PUBLIC_SUPABASE_KEY || '',
```
bvelitchkine, line 14: Isn't it dangerous? I'm not completely clear on this. I reckon it's to remove typing errors?

apps/linkedin-to-notion/src/contents/notion-oauth-nudge-wrapper-styling.ts
```diff
@@ -10,6 +10,10 @@ export const config: PlasmoCSConfig = {
   console.log('Notion OAuth nudge wrapper styling');
   const wrapper = await waitForElement("div[style*='grid-template-rows: 1fr 1fr;']");
   const originalStyle = wrapper.getAttribute('style');
-  const newStyle = originalStyle.replace('grid-template-rows: 1fr 1fr;', 'grid-template-rows: 24px 1fr 1fr;');
+  const newStyle = originalStyle?.replace('grid-template-rows: 1fr 1fr;', 'grid-template-rows: 24px 1fr 1fr;');
+  if (!newStyle) {
```
bvelitchkine, line 14: Test should be on `originalStyle` no?

apps/linkedin-to-notion/src/contents/scrapers/linkedin-data-cleaners.ts
```diff
@@ -22,5 +22,5 @@ export const splitName = (fullName: string): { firstName?: string; lastName?: st
  * @returns a clean job title
  */
 export const cleanJobTitle = (jobTitle: string): string => {
-  return jobTitle.split(/ at| @| chez/g)[0].trim();
+  return jobTitle.split(/ at| @| chez/g)[0]?.trim() || jobTitle;
```
bvelitchkine, line 25: I don't understand why the "or" condition is necessary. If there's no separator in the string, it will still return something coherent. For instance:
```js
"Growth Engineer".split('|')[0]?.trim()
// => "Growth Engineer"
```
Is it to remove typing errors again?

apps/linkedin-to-notion/src/components/LinkedInNotionSidePanel/Form/Form.tsx
```diff
@@ -95,29 +95,35 @@ export const Form = ({
   }, [displayNotionValues, currentNotionValues, linkedinValues]);
 
   useEffect(() => {
-    if (selectedNotionDatabase) {
-      findProfileInNotionDatabase(selectedNotionDatabase, linkedinUrl);
+    if (notionToken && selectedNotionDatabase) {
+      if (linkedinUrl) {
```
bvelitchkine: Why not a triple "and" clause?

apps/linkedin-to-notion/src/background/index.ts
```diff
@@ -4,30 +4,43 @@ export {};
...
   }
 
-  const isOnLinkedIn = tab?.url?.match(linkedInURLRegex);
-  const isOnLinkedInProfile = tab?.url?.match(linkedInProfileURLRegex);
+  if (!tab?.url) {
+    return;
+  }
+
+  const isOnLinkedIn = linkedInURLRegex.test(tab.url);
+  const isOnLinkedInProfile = linkedInProfileURLRegex.test(tab.url);
```
bvelitchkine, line 20: 🤩

apps/linkedin-to-notion/src/background/index.ts
```diff
@@ -4,30 +4,43 @@ export {};
...
+      await chrome.tabs.sendMessage(tabId, 'closeSidePanels');
+    } catch (e) {
+      console.log('Error updating side panels', e);
+    }
+  }
+
   // We update the side panel's content when we are on a LinkedIn profile
   if (isOnLinkedInProfile) {
     try {
+      await chrome.tabs.sendMessage(tabId, 'openLinkedInNotionSidePanel');
```
bvelitchkine: I think you should delete this line. cf. my upcoming PR, we shouldn't rely on content script messaging to open the side panel imo. There's already a matching condition for that in the content script

bvelitchkine (changes requested):
The only blocking comment I have is the message to the content script to reopen the side panel. Apart from that: all clear!

apps/linkedin-to-notion/src/contents/scrapers/linkedin-data-cleaners.ts
GauvainThery: It is and it's necessary. Split can return undefined => [0] of undefined is an error => your app crashes 😬

apps/linkedin-to-notion/src/components/LinkedInNotionSidePanel/Form/Form.tsx
GauvainThery: Good question. Honestly I spent no time on this, I can update.

apps/linkedin-to-notion/src/background/index.ts
GauvainThery: Well, after working on this I tend to disagree on the "we shouldn't rely on content script messaging". Those work perfectly and are super handy. We were just using it wrong

apps/linkedin-to-notion/core/supabase.ts
GauvainThery: Absolutely not 😉 
As dangerous as having just the value as it was. Env variables can be undefined, that's normal. But they'll never be 🆗

bvelitchkine (approved):
LGTM 🎉

